PR #1, opened 2018-01-03
RFC: Schema evolution: field operations

arnau:
### Context

Currently, the Registers Data Model has no way to evolve the data structure
described in the schema and used in data items. This means a Register has to
be designed upfront with the right fields forever given it is an append-only
log. We need a way to be able to evolve the schema definition and a way to
reflect such evolution on existing and new data.

This RFC proposes a way to add and subtract fields from a Register, and
defines the meaning of these operations.


arnau:
This PR will be split in two RFCs.

PR #5, opened 2018-01-26
Add Field (Schema evolution)

arnau:
### Context

Currently, the Registers Data Model has no way to evolve the data structure
described in the schema and used in data items. This means a Register has to
be designed upfront with the right fields forever given it is an append-only
log. We need a way to be able to evolve the schema definition and a way to
reflect such evolution on existing and new data.

This RFC proposes a way to add fields to a Register, and defines the meaning of this operation.

### DIFF
--- a/content/schema-evolution-field-operations/index.md
+++ b/content/schema-evolution-field-operations/index.md
@@ -1,0 +1,194 @@
+---
+rfc:
+start_date: 2017-11-17
+status: draft
+---
+
+# Schema Evolution: Add field operation
+
+## Summary
+
+This RFC proposes a way to add fields to a Register, and defines the meaning
+of this operation. Thus, it doesn’t attempt to cover other field operations
+such as field subtraction or field transformation (e.g.  change the datatype
+of a field).
+
+Note that you can find the [glossary](#glossary) of concepts used throughout
+this RFC at the end of it.
+
+Note: The pseudocode in this RFC resembles [Elm](http://elm-lang.org) but it
+is not intended to be valid.
+
+
+## Motivation
+
+Currently, the Registers Data Model has no way to evolve the data structure
+described in the schema and used in data items. This means a Register has to
+be designed upfront with the right fields forever given it is an append-only
+log. We need a way to be able to evolve the schema definition and a way to
+reflect such evolution on existing and new data.
+
+
+## Explanation
+
+The proposal is to allow adding new field definitions to the schema. As a
+result new items added to the Register will be validated against the new
+schema. To implement this behaviour this RFC defines the mechanism to [add a
+field to the schema](#field-addition) and the mechanism to [add an item to the
+log](#item-addition).
+
+Note: The explanation tries to avoid being prescriptive in how the Register is
+implemented and because of that there is no explicit reference to entries and
+how they participate in either mechanism.
+
+
+### <span id="field-addition">Field addition</span>
+
+The addition of a field, [`Schema.add`](#schema-add-function), adds a
+[field](#field) to the [schema](#schema).
+
+
+##### Example: Add a field to the schema
+
+```elm
+s_0 = Schema [("x", Int)]
+s_1 = Schema.add s_0 ("y", Int) -- s_1 == Schema [("x", Int), ("y", Int)]
+```
+
+
+### <span id="item-addition">Item addition</span>
+
+The item addition operation, [`Items.add`](#items-add-function), adds the item
+if [`Item.validate`](#item-validate-function) succeeds. This RFC doesn't
+change its behaviour, the only thing to keep in mind is that the validation
+function checks against the current schema.
+
+The addition of an item steps are:
+
+1. Validate the incoming item `a` against the current schema `s`: `Item.validate s a`.
+    * If the item is invalid, reject the operation.
+2. Canonicalise the item `a`: `a_c = Item.canonicalise a`.
+3. Add item `a_c` to the Register.
+
+
+### REST API Implications
+
+The REST API has no changes given that it is a read-only API. From the **API
+consumer** perspective, nothing changes but let me make it explicit here.
+
+One cornerstone assumption in the Registers' API is that a consumer is
+expected to handle graciously any Record/Item response that doesn't match
+exactly their knowledge of the schema.
+
+Scenario:
+
+Give a consumer that knows about a schema such as `Schema [("x", Int),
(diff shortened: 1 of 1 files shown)


content/schema-evolution-field-operations/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+1. Validate `a` against the current schema `s`: `Item.validate s a`.
+    * If the item is invalid, reject the operation.
+2. Canonicalise the item `a_p`: `a_c = Item.canonicalise a_p`.
+3. Add item `a_c` to the Register.
+
+
+### REST API Implications
+
+The REST API has no changes given that it is a read-only API.
```
nacnudus: Can it cope with old items that don't have the new field?

arnau: Yes, the assumption in a Register (previous to this RFC) is that a consumer has to be able to handle any response serialised in the requested format (e.g. json) and expect to receive part of the known fields. Also it has to ignore unknown fields. That is:

Given a consumer that knows about a schema such as `Schema [("x", Number), ("y", Number), ("z", String)]` and assuming `x` is the primary field, a consumer has to be able to expect and handle:

#### All known fields present

Given `{"x": 1, "y": 2, "z": "foo"}`, keep `{"x": 1, "y": 2, "z": "foo"}`.

#### Some known fields present

Given `{"x": 1, "y": 2}`, keep `{"x": 1, "y": 2}` and not fail because the missing `z`.

#### Some unknown fields present

Given `{"x": 1, "y": 2, "z": "foo", "n": 42}`, keep `{"x": 1, "y": 2, "z": "foo"}` and not fail because the unknown `n`.

Arguably in the last case, the consumer may want to check if there is a new schema and act accordingly but it depends on the implementation.

arnau: When the serialisation has to keep columnar consistency (e.g. CSV), then items previous to the new field would have a missing value there. The meaning of this missing value is the `null` as "unknown".

This RFC doesn't add any mechanism for defining default values other than `null` because it is a complex topic on its own that deserve full attention to be sure we are choosing the right tradeoffs.

karlbaker02: For clarification on the last point of **Some unknown fields present**, are we talking about the consumer reading from the registers API here or adding to the register (I'm also the former from the remainder of the description, but a little confused here).

Given `{"x": 1, "y": 2, "z": "foo", "n": 42}`, the current Java implementation of the app would disallow the Item because of an unknown field for the register, which here is `n`. Conversely, if this Item were already part of the register and the consumer was reading from the register when field `n` was not part of the register, then this would be a case of field removal rather than addition.

The current Java implementation may just be that, and it may be perfectly acceptable for a register to accept an incoming Item with an unknown field, but just want to double-check this here. Apologies if I've missed the point!

arnau: @karlbaker02 the points I was talking about are from the point of view of the API consumer (which is read-only at all effects). So, from the point of view of the current Java implementation (API provider) `n` would be known and hence the item valid. I'm not following the "field removal" part I'm afraid.

arnau:
I added to the RFC the clarifications I made here. It seems appropriate to be more explicit and concrete about what are the expectations for robust API consumer implementations. In turn, the explanation will land to the specification and perhaps to the API documentation. @bahmady the "REST API Implications" section might be relevant for you once this RFC gets approved.

content/schema-evolution-field-operations/index.md
```diff
@@ -0,0 +1,194 @@
...
+### <span id="item-addition">Item addition</span>
+
+The item addition operation, [`Item.add`](#item-add-function), adds the item
+if [`Item.validate`](#item-validate-function) succeeds. This RFC doesn't
+change its behaviour, the only thing to keep in mind is that the validation
+function checks against the current schema.
+
+The addition of an item steps are:
+
+1. Validate `a` against the current schema `s`: `Item.validate s a`.
```
michaelabenyohai: It would be good to make it clear that `a` is an item at this point.

content/schema-evolution-field-operations/index.md
```diff
@@ -0,0 +1,194 @@
...
+* **All known fields are present**
+
+  Given `{"x": 1, "y": 2, "z": "foo"}`, keep `{"x": 1, "y": 2, "z": "foo"}`.
+
+* **Some known fields are present**
+
+  Given `{"x": 1, "y": 2}`, keep `{"x": 1, "y": 2}` and ignore the fact that
+  `z` is missing.
+
+* **Some unknown fields present**
```
michaelabenyohai: I agree that this will work for field addition - that "a consumer is expected to handle graciously a response that doesn't match their knowledge of the schema". However, especially for this last case, I bet there are some people that haven't integrated in this way, and are simply consuming what we give them in the item. Do you feel that this is a bit "new" and we need to make it really obvious in our API documentation that we expect this?

arnau: For my conversations with Paul, this is the intended way since the beginning. That said, I would like the API docs to be much more clear about this expectations.

Now, if someone prefers accepting any field found in an item because they don't use the schema and are happy to handle data in a different way, it's ok. The point is that they shouldn't get an exception because of this.

content/schema-evolution-field-operations/index.md
```diff
@@ -0,0 +1,194 @@
...
+If the field already exists the operation behaves like the identity function
+so:
+
+```elm
+schema' = Schema.add schema field -- schema' == schema
+```
+
+### <span id="item-add-function">Item `add` function</span>
+
+The function `Item.add` takes a list of items, a schema and an item and
```
michaelabenyohai: Should this be `Items.add` (plural)? I think it's meant to be the same as the example below which is plural? The title of the section also uses Item singular but would be good to have all three consistent.

content/schema-evolution-field-operations/index.md
arnau: Given that we don't know for sure all current implementations, I'm happy to start by putting a clear explanation about expectations in the API docs (sooner rather than latter) and put a expected data for when this will be used in our end.

content/schema-evolution-field-operations/index.md
arnau: Ah yes, @karlbaker02 pointed it out as well and forgot to fix it 👍

arnau:
Final rewrite to squash all intermediate commits.

PR #10, opened 2018-03-08
Add URI Consolidation

arnau:
### Context

This RFC proposes a change in the REST API to consolidate endpoints of the
same resource type to use the same naming convention so CURIEs can be used to
refer to the full collection of records within a Register.

### Guidance to review

Have a read and try to find corner cases where the proposal might break or create inconsistencies.

### DIFF
--- a/content/uri-consolidation/index.md
+++ b/content/uri-consolidation/index.md
@@ -1,0 +1,64 @@
+---
+rfc:
+start_date: 2018-03-08
+pr:
+status: draft
+---
+
+# URI Consolidation
+
+## Summary
+
+This RFC proposes a change in the REST API to consolidate endpoints of the
+same resource type to use the same naming convention so CURIEs can be used to
+refer to the full collection of records within a Register.
+
+
+## Motivation
+
+Currently the REST API uses the plural form of a resource type for collection
+endpoints and the singular form for collection elements. Based on how Registers
+use CURIEs to reference records this design choice makes it impossible to
+refer to the whole list of records using the same CURIE prefix.
+
+
+## Explanation
+
+To address this limitation, we have two options: use the same form for the
+endpoints, or define two prefixes per register. Using the latter option would
+make defining and evolving registers more complex than using the former
+option, so we propose to only use the plural form for endpoints.
+
+We selected the plural form because it's more common in well-known APIs (e.g.
+[GitHub API v3](https://developer.github.com/v3)).
+
+Endpoint mappings:
+
+* From `/record/{field-value}` to `/records/{field-value}`.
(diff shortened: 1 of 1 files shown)


content/uri-consolidation/index.md
```diff
@@ -0,0 +1,63 @@
...
+
+Endpoint mappings:
+
+* From `/record/{field-value}` to `/records/{field-value}`.
+* From `/entry/{entry-number}` to `/entries/{entry-number}`.
+* From `/item/{item-hash}` to `/items/{item-hash}`.
+
+Note that we have not yet implemented `/items`, but it is in the pipeline.
+
+The `/records/{field-value}` change is the key change that enables defining a
```
nacnudus: "key" is a loaded term in this context.  How about "principal" instead?

content/uri-consolidation/index.md
```diff
@@ -0,0 +1,63 @@
...
+
+We selected the plural form because it's more common in well-known APIs (e.g.
+[GitHub API v3](https://developer.github.com/v3)).
+
+Endpoint mappings:
+
+* From `/record/{field-value}` to `/records/{field-value}`.
+* From `/entry/{entry-number}` to `/entries/{entry-number}`.
+* From `/item/{item-hash}` to `/items/{item-hash}`.
+
```
michaelabenyohai: We could also add the below. It's not documented yet but the endpoint exists in the implementation and for two beta registers.

From `/index/{index-name}/record/{key}` to `/index/{index-name}/records/{key}`

content/uri-consolidation/index.md
```diff
@@ -0,0 +1,63 @@
...
+endpoints, or define two prefixes per register. Using the latter option would
+make defining and evolving registers more complex than using the former
+option, so we propose to only use the plural form for endpoints.
+
+We selected the plural form because it's more common in well-known APIs (e.g.
+[GitHub API v3](https://developer.github.com/v3)).
+
+Endpoint mappings:
+
+* From `/record/{field-value}` to `/records/{field-value}`.
```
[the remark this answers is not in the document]
arnau: Yes, I just used the same convention as we currently have in the spec and api docs.  But I agree it would be better as something else.

michaelabenyohai:
Looks great other than minor comments.

michaelabenyohai:
Actually, also need to add:

From `/record/{field-name}/entries` to `/records/{field-name}/entries`

content/uri-consolidation/index.md
arnau: As discussed we will address this when it's formally introduced to the model and API.

PR #11, opened 2018-03-29
RFC0005: Reference strategy

arnau:
### Context

This RFC proposes consolidating references in Registers to CURIEs only. 

### Guidance to review

* Review the description of the reality for foreign keys and CURIEs is accurate.
* Review the proposal is coherent.
* Raise any doubts or concerns.

### DIFF
--- a/content/reference-strategy/index.md
+++ b/content/reference-strategy/index.md
@@ -1,8 +1,9 @@
 ---
 rfc: 0005
 start_date: 2018-03-29
+decision_date: 2018-08-13
 pr: openregister/registers-rfcs#11
-status: draft
+status: approved
 ---
 
 # Reference strategy
--- a/content/reference-strategy/index.md
+++ b/content/reference-strategy/index.md
@@ -1,7 +1,7 @@
 ---
-rfc:
+rfc: 0005
 start_date: 2018-03-29
-pr:
+pr: openregister/registers-rfcs#11
 status: draft
 ---
 
--- a/content/reference-strategy/index.md
+++ b/content/reference-strategy/index.md
@@ -13,9 +13,10 @@ This RFC proposes CURIEs as the only way to reference between records and from a
 to a register (set of records). It can be seen as the natural consequence of
 [RFC 0002: URI Consolidation](https://github.com/openregister/registers-rfcs/blob/master/content/uri-consolidation/index.md).
 
-Note that other types of references (e.g. subset of records) are out of the
-scope. See the discussion [Group References](https://github.com/openregister/registers-rfcs/issues/4)
-for more on this.
+Note that other types of references (e.g. subset of records) are out of scope.
+See the discussion [Group
+References](https://github.com/openregister/registers-rfcs/issues/4) for more
+on this.
 
 ## Motivation
 
@@ -28,33 +29,43 @@ register (including itself). You have to deduce by the name of the field that
 the values are in fact identifiers (primary keys) for records in a Register of
 the same name. To get the URI of the referenced resource you have to:
 
-1. Given a fieldname `country`
+1. Given a field with name `x`, where the record for `x` in the Field register
+   has the `register` field populated with value `country`
 1. And a value `GB`
 1. Compose a register subdomain: `country.register.gov.uk`
 1. Compose a record set URI: `https://country.register.gov.uk/records/`
 1. Concatenate the URI with the value: `https://country.register.gov.uk/records/GB`
 
-There is a special case here where a foreign key refers to a register register
-record. Then the meaning of the relation is “everything in the register
-described by the linked record”.
+There is a case that can be seen as a special case where a foreign key refers
+to a register register record. Then the meaning of the relation can be seen as
+“everything in the register described by the linked record”.
 
 1. Given a fieldname `register`
 1. And a value `country`
-1. Compose a register subdomain: `country.register.gov.uk`
-1. Compose a set URI: `https://country.register.gov.uk/records/`
+1. Compose a register subdomain: `register.register.gov.uk`
+1. Compose a record set URI: `https://register.register.gov.uk/records/`
+1. Concatenate the URI with the value: `https://register.register.gov.uk/records/country`
+1. Derive the record set URI: `https://country.register.gov.uk/records/`
+
+The last step is not specified anywhere so it is an internal convention more
+than anything a regular user might know. It is defined here in order to mirror
+the capabilities for CURIEs.
 
 #### Benefits
 
 * Potentially familiar to people familiar with the Relational Data Model.
 * Hints a single type of thing for any value in the field.
 
-#### Problems
+#### Limitations
 
-* To know a field contains a foreign key, you need to look up the register field
-in the field definition. If it is informed, the datatype is not “string” but
-“key”.
+* To know a field contains a foreign key, you need to look up the `register`
+  field in the field register. If it is informed, the original field is a “key”
+  of type defined in the `datatype` field.
 * Only able to reference one register per field.
-* Linking a full set of records is a special case.
+* You need a field for each register linked to even if they have the same type
+  of relationship with the record (e.g. owned by, managed by).
+* Linking a full set of records is not possible. Although there is a
+  convention that could be standarised previous section.
 
 
 ### CURIE
@@ -89,7 +100,7 @@ step:
 * It's built on a existing open standard. Technically it is a note and not a
   recommendation but it is used in many standards, including RDF.
 
-#### Problems
+#### Limitations
 
 * A consumer has to know the assumptions that lead from a prefix to a URI. And
   know it's not the standard way to work with CURIEs.
@@ -132,10 +143,24 @@ With this context, the following expansions are expected to be true:
 Note that currently the context is managed partially by the Register register.
 Further work is expected to evolve it to match this definition of context.
 
-## Conclusion
+## Naming conventions
 
-This RFC accepts CURIE as the only mechanism for linking between register from
-now on given they offer the flexibility required by registers.
+Foreign keys have been using the same field definition as the primary key in
+the targeted register. This means that a sigle entry in the field register
+defines a field used in one register as the primary key and in another
+register as a foreign key. This convention derives from another one where the
+primary key field is always the same as the register identifier. E.g. the
+register with id `country`, has a field `country` that acts as the primary key
+for that register. Another register with a field `country` will use the field
+as a foreign key.
 
-Existing registers that use foreign keys will be maintained to avoid
-disruption in the registers ecosystem.
+With CURIEs this assumption weakens. A future RFC will define the new
+convention for naming fields of type CURIE although this will be _informative_
+not _normative_. The current thinking is to move away from using names and
+start using relationships. For example, given a register with a link to the
+organisation that manages the resource, it would use something like
+`managed-by` instead of `manager` or `organisation`.
+
+A future RFC will formalise the way to explicitly identify the primary key for
+a register so the convention of using the register identifier will be
+addressed there.
--- a/content/reference-strategy/index.md
+++ b/content/reference-strategy/index.md
@@ -1,0 +1,141 @@
+---
+rfc:
+start_date: 2018-03-29
+pr:
+status: draft
+---
+
+# Reference strategy
+
+## Summary
+
+This RFC proposes CURIEs as the only way to reference between records and from a record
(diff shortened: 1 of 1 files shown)


content/reference-strategy/index.md
```diff
@@ -0,0 +1,135 @@
...
+the values are in fact identifiers (primary keys) for records in a Register of
+the same name. To get the URI of the referenced resource you have to:
+
+1. Given a fieldname `country`
+1. And a value `GB`
+1. Compose a register subdomain: `country.register.gov.uk`
+1. Compose a record set URI: `https://country.register.gov.uk/records/`
+1. Concatenate the URI with the value: `https://country.register.gov.uk/records/GB`
+
+There is a special case here were a foreign key refers to a register register
```
nacnudus: Typo, `were` should be `where`

content/reference-strategy/index.md
```diff
@@ -0,0 +1,135 @@
...
+With this context, the following expansions are expected to be true:
+
+* `country:GB` -> `https://country.register.gov.uk/records/GB`
+* `country:ZZ` -> `https://country.register.gov.uk/records/ZZ`
+* `lae:BIR` -> `https://local-authority-eng.register.gov.uk/records/BIR`
+* `example:32` -> `https://example.org/32`
+
+
+TODO: Where does the context live? It can't be part of the register API
+because a Register could be part of multiple environments.
```
nacnudus: What is an example of a register being part of multiple environments?

content/reference-strategy/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+With this context, the following expansions are expected to be true:
+
+* `country:GB` -> `https://country.register.gov.uk/records/GB`
+* `country:ZZ` -> `https://country.register.gov.uk/records/ZZ`
+* `lae:BIR` -> `https://local-authority-eng.register.gov.uk/records/BIR`
+* `example:32` -> `https://example.org/32`
+
+
+TODO: Where does the context live? It can't be part of the register API
```
nacnudus: It needs to be tamper-proof though.

content/reference-strategy/index.md
arnau: The immediate situation would be a Register being part of the register.gov.uk environment and being part of a local/test environment.

Another one, artificial case right now, would be a register from NHS that is part of the NHS environment and part of ours.

content/reference-strategy/index.md
arnau: What do you mean?

nacnudus: If someone tampers with the context, then a CURIE would resolve to the wrong data.

content/reference-strategy/index.md
nacnudus: Does that mean the CURIE doesn't resolve to a "single source of truth" but instead just one of many places where that truth is published?

content/reference-strategy/index.md
arnau: Ah yes, it's something to keep an eye on.

content/reference-strategy/index.md
arnau: It resolves to the source of truth you decided. Which most of the time is our environment.

arnau: @michaelabenyohai and I had a chat about another topic that led to a discussion around the possibility to encode more specific data instead of the CURIE (as per the spec) so it could be proved correct.  But right now it is just a thought.

arnau:
> Specifically, we still have the rule that the key field of a register must have the same name as the register. This guidance instructs that these fields must no longer be created as a foreign key (which they always have previously). This in turn means that this field name can never be reused as a link from another register (which we currently do a lot). Instead, we will have to create a new field with a slightly different name, so could end up proliferating fields that come in "pairs" with similar-but-not-quite-the-same name.
>
>I feel we could treat this as the key field should describe "the thing", whereas the curie field in the "other thing" register should describe the relationship of "the thing" to "other thing". In practice I think this could be hard though.

Very good point, I'll address it in the RFC.

content/reference-strategy/index.md
```diff
@@ -0,0 +1,142 @@
...
+
+* Potentially familiar to people familiar with the Relational Data Model.
+* Hints a single type of thing for any value in the field.
+
+#### Problems
+
+* To know a field contains a foreign key, you need to look up the register field
+in the field definition. If it is informed, the datatype is not “string” but
+“key”.
+* Only able to reference one register per field.
```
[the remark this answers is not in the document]
arnau: You are right in that it is not a problem in itself, just a restriction. The problem derived from this restriction is the need to have multiple fields for similar things when the situation arises. I'll think in a better way to phrase this.

content/reference-strategy/index.md
```diff
@@ -0,0 +1,142 @@
...
+1. Compose a set URI: `https://country.register.gov.uk/records/`
+
+#### Benefits
+
+* Potentially familiar to people familiar with the Relational Data Model.
+* Hints a single type of thing for any value in the field.
+
+#### Problems
+
+* To know a field contains a foreign key, you need to look up the register field
```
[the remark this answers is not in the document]
arnau: Yes, my point was around not being able to know by the datatype.

content/reference-strategy/index.md
```diff
@@ -0,0 +1,142 @@
...
+
+Note that currently the context is managed partially by the Register register.
+Further work is expected to evolve it to match this definition of context.
+
+## Conclusion
+
+This RFC accepts CURIE as the only mechanism for linking between register from
+now on given they offer the flexibility required by registers.
+
+Existing registers that use foreign keys will be maintained to avoid
```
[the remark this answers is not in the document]
arnau: We should, yes.

content/reference-strategy/index.md
```diff
@@ -0,0 +1,142 @@
...
+scope. See the discussion [Group References](https://github.com/openregister/registers-rfcs/issues/4)
+for more on this.
+
+## Motivation
+
+Currently, there are two mechanisms for referencing: a foreign key and a CURIE.
+
+### Foreign Key
+
+A foreign key is a regular string that _happens_ to be an identifier in
```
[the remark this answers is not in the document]
arnau: Thinking about this, it connects with another discussion about expectations around primary key values: #18 

In there I was only considering strings but your comment opens the door to more (potential) complexity.

I think we care about how flexible we must be with identifiers that can be used in CURIEs (and because historical reasons in foreign keys). In particular with CURIEs, the value needs to be encoded in terms of a URL path so `foo:1` would be the same if the identifier is a string or an integer. I'm leaning towards exploring issues derived to restrict identifiers to be only a subset of UTF-8.

content/reference-strategy/index.md
```diff
@@ -0,0 +1,142 @@
...
+another register. You have to query the Field register and check the
+`register` field.  If it is informed, it is a primary key in that register and
+when found in another register it acts as a foreign key. To get the URI of the
+referenced resource you have to:
+
+1. Given a fieldname `country`
```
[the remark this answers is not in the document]
arnau: Do you know of any cases where this is not the case?

content/reference-strategy/index.md
```diff
@@ -0,0 +1,142 @@
...
+1. And a value `GB`
+1. Compose a register subdomain: `country.register.gov.uk`
+1. Compose a record set URI: `https://country.register.gov.uk/records/`
+1. Concatenate the URI with the value: `https://country.register.gov.uk/records/GB`
+
+There is a special case here where a foreign key refers to a register register
```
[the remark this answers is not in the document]
arnau: I think I'll need a chat with you on this one.

arnau:
@michaelabenyohai I've changed a few bits to address some of your comments, please have another read and see if it's better now.

MatMoore:
The proposal makes sense to me based on what I know about how things work now.

Am I right in thinking that to actually replace the existing foreign keys with curies we first need to create the [metadata log](https://github.com/openregister/registers-rfcs/pull/20)?

Also, even if we're not adding any more foreign keys, will we retain any information in the spec about the semantics of them, or will it become purely an implementation detail of specific ORJ registers? I'm just wondering whether a spec-compliant client would still be expected to resolve these kinds of links.

arnau:
> Am I right in thinking that to actually replace the existing foreign keys with curies we first need to create the metadata log?

Yes, in particular the schema evolution (adding fields) #5 

> Also, even if we're not adding any more foreign keys, will we retain any information in the spec about the semantics of them, or will it become purely an implementation detail of specific ORJ registers? I'm just wondering whether a spec-compliant client would still be expected to resolve these kinds of links.

Good point, I think the spec should, at best, mention that at some point a register had this assumption; so some sort of informative section. But a client shouldn't be expected to implement legacy assumptions.

Moreover, the current spec doesn't define foreign keys at all, so in a way we are addressing an assumption we have encoded in our implementation and clients.

content/reference-strategy/index.md
```diff
@@ -0,0 +1,173 @@
...
+addressed there.
+
+## Conclusion
+
+This RFC accepts CURIE as the only mechanism for linking between register from
+now on given they offer the flexibility required by registers.
+
+Existing registers that use foreign keys will be maintained to avoid
+disruption in the registers ecosystem. An independent task will define and
+execute the migration from foreing keys to CURIEs.
```
MatMoore: there's a typo here: "foreing"

MatMoore: actually you could just drop this whole sentence as it doesn't say when this will happen or how.

content/reference-strategy/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+# Reference strategy
+
+## Summary
+
+This RFC proposes CURIEs as the only way to reference between records and from a record
+to a register (set of records). It can be seen as the natural consequence of
+[RFC 0002: URI Consolidation](https://github.com/openregister/registers-rfcs/blob/master/content/uri-consolidation/index.md).
+
+Note that other types of references (e.g. subset of records) are out of the
```
MatMoore: "out of the scope" should be "out of scope"

arnau:
> The limitation of not knowing how to resolve a CURIE to a URL could be solved by including the context within the registers resource.

This is indeed the current thinking for how to handle reference resolution. It requires schema evolution to add this information to the catalog (register register).

But it's not of this RFC to say how the catalog works. There are more twists on this, happy to discuss it further.

content/reference-strategy/index.md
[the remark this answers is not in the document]
michaelabenyohai: The field [`allergen-group`](https://field.register.gov.uk/records/allergen-group) is one and the field [`fields`](https://field.register.gov.uk/records/fields) is another.

content/reference-strategy/index.md
```diff
@@ -34,27 +34,36 @@ the same name. To get the URI of the referenced resource you have to:
...
-* To know a field contains a foreign key, you need to look up the register field
-in the field definition. If it is informed, the datatype is not “string” but
-“key”.
+* To know a field contains a foreign key, you need to look up the `register`
+  field in the field register. If it is informed, the original field is a “key”
+  of type defined in the `datatype` field.
 * Only able to reference one register per field.
-* Linking a full set of records is a special case.
+* You need a field for each register even if they have the same type of
+  relationship with the record (e.g. owned by, managed by).
```
michaelabenyohai: I don't understand this point.

content/reference-strategy/index.md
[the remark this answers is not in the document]
michaelabenyohai: I think this point still needs addressing.

content/reference-strategy/index.md
arnau: At all or part of it? Let me try again:

With foreign keys, the field has a local identifier so the field itself is dedicated to a single type of link (e.g. local-authority-eng). If you need to provide links to e.g. local-authority-nir as well, you need another column. Even though both have the same intention (relationship wise).

content/reference-strategy/index.md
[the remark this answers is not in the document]
arnau: To clarify, this plan is out of scope for this RFC.

content/reference-strategy/index.md
michaelabenyohai: Yes that makes sense now. I think it would be good to clarify that you "need a field for each register linked to"

content/reference-strategy/index.md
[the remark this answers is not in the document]
arnau: Ok, changed with your suggestion.

PR #12, opened 2018-04-05
RSF version 1

arnau:
### Context

RSF has been documented in multiple places but never in a consistent manner. This is an attempt to describe the current state of things and serve as the starting point to create the first RSF specification.

Note that RSF has no versioning system. This is a good opportunity to discuss if we should add one to allow non backwards-compatible changes without breaking clients.

### Guidance to review

Review the current implementation of RSF is accurately described in this RFC and that the explanation is understandable and unequivocal.


content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+
+### REST API
+
+TODO
+
+The current implementation uses `GET /download-rsf`. The main issue with that
+is that diverges from the rest of the API where serialisation is expressed
+either via suffix or via media type. The problem with using the same approach,
+say `GET /register.rsf` is that we are not providing the same information when
+querying `GET /register.json`.
```
nacnudus: Does this mean that the endpoint `GET /download-rsf` doesn't provide the same information as `GET /register.rsf`, or that the user has to provide different information?

```text
The current implementation uses `GET /download-rsf`. The main issue with that
is that diverges from the rest of the API where serialisation is expressed
either via suffix or via media type. The problem with using the same approach,
say `GET /register.rsf` is that we are not providing the same information when
querying `GET /register.json`.
```

content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+
+What is a good name for a resource that represents the whole raw database?
```
nacnudus: I think "register" is a good name for a register ;)

`GET /register.rsf`

content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+GET /{db resource}.jsonl
+
+GET /{db resource}.jsonl
+Accept: application/x-ndjson
+```
+
+
+### Commands
+
+#### <a id="assert-root-hash-command">`assert-root-hash` command</a>
```
nacnudus: Why are there two `assert-root-hash` commands per register?  How do you use them?

content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+For example, the empty root hash:
+
+```
+assert-root-hash	sha-256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+```
+
+#### <a id="add-item-command">`add-item` command</a>
+
+Adds a new [Item resource][item-res] to the register. It will require an
+[`append-entry` command](#append-entry-command) to make it visible to users.
```
nacnudus: Instead of "make it visible to users", I would put "associate the item with a record key".

content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+Appends a new [Entry resource][entry-res] to the register.
+
+##### Arguments
+
+1. The `type` of the entry determines if the entry belongs to the data log
+   (`user`) or to the metadata log (`system`).
+2. The `key` of the entry. The primary key field is the field with the same
+   name as the register.
+3. The `timestamp` of the entry. This is the time at which the entry was
+   appended to the register.
```
nacnudus: I believe this is the first time we have defined what the `timestamp` means.  Are we sure?

content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+### Validation rules
+
+A RSF list of commands is expected to conform to the following rules:
+
+* [Commands](#commands) are executed in order of appearance, top to bottom.
+* Entries are numbered in sequence in order of appearance starting with 1 if
+  the register is empty, otherwise incrementing on the latest entry number
+  found in the register.
```
nacnudus: Is there any way to tell whether an RSF file represents a complete register, or only appends to a register?

content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+* An [`append-entry` command](#append-entry-command) must always appear after
+  the [`add-item` command](#add-item-command) that introduces the item is
+  referencing.
+* It is illegal to have orphan items. An `add-item` must have at least one
+  `append-entry` referencing to the item.
```
nacnudus: If the the validity of an `add-item` depends on a subsequent `append-entry`, then a file that gets accidentally truncated in transit will be invalid.  This seems unnecessarily strict, but I haven't thought hard about the disadvantages of orphan items.  Is it to do with treating patches as transactions?

content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+* It is illegal to have broken references. An `append-entry` must reference an
+  item previously introduced by an `add-item` command.
+* The item in the `add-item` command must always be in the canonical form.
+
+
+#### Type checking
+
+Although not part of the RSF specification, it is worth mentioning that a
+Registers' implementation is expected to type check the data according to the
```
nacnudus: I don't think the apostrophe is needed on "Registers'"

content/rsf-spec/index.md
```diff
@@ -0,0 +1,263 @@
...
+* The `official-name` field is a String.
+
+Each datatype must be parsed according to the [datatype specification][datatype-spec].
+
+A RSF patch (set of commands) must be treated as a single transaction. If
+there is a validation error, the whole patch must be rejected and any changes
+to the state rolled back.
+
+
+### Examples
```
nacnudus: Should these examples be valid patches?

content/rsf-spec/index.md
arnau: Well, `/register.json` right now offers some sort of summary-metadata of the register. RSF by nature is the full register. So both things are at odds.

nacnudus: I didn't know about that endpoint :grimacing:.  It goes without saying that I think it should have a different name, e.g. `/summary.json`.

content/rsf-spec/index.md
arnau: `assert-root-hash` can appear as many times as you want in a RSF file. The idea is that every time you find one, you can use it to verify you have got all the previous entries right, it's a checkpoint really.

content/rsf-spec/index.md
arnau: umm, I like what you are suggesting but I don't like the idea of using the word "record". @michaelabenyohai ideas?

content/rsf-spec/index.md
arnau: I think it's the first time we have something that clear yes. Based on usage I'd say yes, the timestamp is the consequence of _minting_ an item so it's the recording time for the entry. It mimics git's behaviour.

content/rsf-spec/index.md
arnau: The only way as far as I know is via `assert-root-hash`. In a situation where the RSF has no assertions you wouldn't know.

content/rsf-spec/index.md
arnau: Yes, the original discussion about tradeoffs of allowing orphan items or not made the team decide to be transactional per file so if something wrong happens in the middle of consuming a delta, it must be rolled back so you can retry again with the same delta without complex gymnastics. There were other considerations around how to handle orphans in the API and the garbage collection of them if allowed.

I think it's a reasonable restriction given the consequences of not having it.

content/rsf-spec/index.md
arnau: They should yes, if I haven't mess it up, they are 😱

content/rsf-spec/index.md
nacnudus: Thanks, that answers my later question about whether an RSF file starts at the beginning of a register.  I should have checked with `echo -n | sha256sum`.  In that case should this spec require that the the first line of an RSF file be `assert-root-hash`?

content/rsf-spec/index.md
nacnudus: Does it matter whether the `add-item` command comes before or after the `append-entry` one?

content/rsf-spec/index.md
arnau: I suggested that in another comment, the potential problem is that currently it is not a requirement so some RSF might be broken if we impose this restriction. Let me assess how big of a deal it is really.

content/rsf-spec/index.md
nacnudus: How about "There must be a corresponding `append-entry` command that refers to the item's hash"?  Noting the later comment on "It is illegal to have orphan items", which insists on the `append-entry` for the file to be valid.

arnau: That could work 👍

content/rsf-spec/index.md
arnau: At least it's clearer in intention 👍

content/rsf-spec/index.md
nacnudus: In that case I think it would be useful to require an assert-root-hash in the first line of every file, for the sake of reassembling a register from its patches.

content/rsf-spec/index.md
arnau: what would be the problem of having timestamps out of sequence? The order of the log is dictated by the entry number.

content/rsf-spec/index.md
nacnudus: Haha, I don't think you've messed them up.  But I think the spec should require an `assert-root-hash` in the first and final lines.

content/rsf-spec/index.md
arnau: It could mess with tooling using the timestamp to infer something related to time outside "time of recording" but agains, nothing you wouldn't see in git or similar.

I thing it's up to the tooling to be zealous about timestamps to the extent a tool can be.

content/rsf-spec/index.md
arnau: If my memory serves me well, it has to do with exposing the entry to the API users and not having the item available yet (the record couldn't be computed either). This is me reading between lines and guessing a bit here. In all fairness if we think it's ok to expose an entry as soon as it gets parsed, it means we are not behaving transactionally (a rollback could cause a disruption in someone that already consumed an item). I'll try to pull out the rationale for this restriction but, if it's not satisfactory I might reconsider having the restriction in place.

content/rsf-spec/index.md
nacnudus: Since the scope of this review is "Review the current implementation of RSF is accurately described in this RFC", I'll back off and open this in an issue instead.

content/rsf-spec/index.md
nacnudus: Okay I'm satisfied :slightly_smiling_face:

content/rsf-spec/index.md
```diff
@@ -0,0 +1,264 @@
...
+log              = command *(CRLF command) [CRLF]
+command          = add-item / append-entry / assert-root-hash
+
+assert-root-hash = %s"assert-root-hash" HTAB hash
+
+add-item         = %s"add-item" HTAB CANONREP
+
+append-entry     = %s"append-entry" HTAB type HTAB key HTAB timestamp HTAB hash-list
+type             = "user" / "system"
+key              = alphanum / %x2D / %x5F
```
michaelabenyohai: I'm not sure I follow what `alphanum / %x2D / %x5F` means. Can you explain?

content/rsf-spec/index.md
```diff
@@ -0,0 +1,264 @@
...
+* RFC5234: `ALPHA` (letters), `CRLF` (carriage return, line feed), `DIGIT`
+  (decimal digits), `HEXDIG` (hexadecimal digits) and `HTAB` (horizontal tab).
+* Registers specification: [`CANONREP`][canon-rep] (canonical representation).
+  Note that, in turn, it depends on [RFC8259](https://tools.ietf.org/html/rfc8259).
+
+```abnf
+log              = command *(CRLF command) [CRLF]
+command          = add-item / append-entry / assert-root-hash
+
+assert-root-hash = %s"assert-root-hash" HTAB hash
```
michaelabenyohai: Does `%s` mean case sensitive? If so, why do we only use it for "add-item" etc and not for things like "user"? Not that I think we've ever specified whether these things are case sensitive or not.

content/rsf-spec/index.md
michaelabenyohai: There is another endpoint called `/download-register` that downloads a ZIP file containing "the whole database". We should also consider whether we keep that an/or rename it in line with these thoughts.

content/rsf-spec/index.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+
+### Commands
+
+#### <a id="assert-root-hash-command">`assert-root-hash` command</a>
+
+Asserts that the provided root hash is the same as the one computed from the
+current entry log as defined in the [Digital Proofs][digital-proofs]
+specification.
```
michaelabenyohai: Should we specify that although the included in RSF, an `assert-root-hash` command currently ignores all `system` entries? It only includes `user` entries.

content/rsf-spec/index.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+#### <a id="append-entry-command">`append-entry` command</a>
+
+Appends a new [Entry resource][entry-res] to the register.
+
+##### Arguments
+
+1. The `type` of the entry determines if the entry belongs to the data log
+   (`user`) or to the metadata log (`system`).
```
michaelabenyohai: This implies that the data log and the metadata log are separate things. Is this intentional? I know they are kind of separate currently (e.g. system entries are ignored in root-hashes) but they do all appear in the same "log" in the RSF. I guess this kind of does correctly explain how things are now (even if we want to change them).

content/rsf-spec/index.md
michaelabenyohai: Also note there are other RSF endpoints:

- `/download-rsf/n` gets RSF for the register after entry-number `n`, i.e. it returns the whole register from entry number `n+1` to the end of the register.
- `/download-rsf/n/m` gets RSF for entries `n+1` to `m`'

content/rsf-spec/index.md
michaelabenyohai: Should we specify that this currently does not include `system` entries (though it should do).

content/rsf-spec/index.md
```diff
@@ -0,0 +1,264 @@
...
+  found in the register.
+* An [`append-entry` command](#append-entry-command) must always appear after
+  the [`add-item` command](#add-item-command) that introduces the item is
+  referencing.
+* It is illegal to have orphan items. An `add-item` must have at least one
+  `append-entry` referencing to the item.
+* It is illegal to have broken references. An `append-entry` must reference an
+  item previously introduced by an `add-item` command.
+* The item in the `add-item` command must always be in the canonical form.
+
```
michaelabenyohai: Should we mention anything about identical consecutive append-entry commands currently being illegal?

i.e. this is illegal
```
append-entry	user	GB	2010-11-12T13:14:15Z	sha-256:08bef0039a4f0fb52f3a5ce4b97d7927bf159bc254b8881c45d95945617237f6
append-entry	user	GB	2010-11-12T13:14:50Z	sha-256:08bef0039a4f0fb52f3a5ce4b97d7927bf159bc254b8881c45d95945617237f6
```

This is ok
```
append-entry	user	GB	2010-11-12T13:14:15Z	sha-256:08bef0039a4f0fb52f3a5ce4b97d7927bf159bc254b8881c45d95945617237f6
append-entry	user	GB	2010-11-12T13:14:30Z	sha-256:490636974f8087e4518d222eba08851dd3e2b85095f2b1427ff6ecd3fa482435
append-entry	user	GB	2010-11-12T13:14:50Z	sha-256:08bef0039a4f0fb52f3a5ce4b97d7927bf159bc254b8881c45d95945617237f6
```

content/rsf-spec/index.md
```diff
@@ -0,0 +1,264 @@
...
+* Entries are numbered in sequence in order of appearance starting with 1 if
+  the register is empty, otherwise incrementing on the latest entry number
+  found in the register.
+* An [`append-entry` command](#append-entry-command) must always appear after
+  the [`add-item` command](#add-item-command) that introduces the item is
+  referencing.
+* It is illegal to have orphan items. An `add-item` must have at least one
+  `append-entry` referencing to the item.
+* It is illegal to have broken references. An `append-entry` must reference an
+  item previously introduced by an `add-item` command.
```
michaelabenyohai: In the current implementation, if a corresponding `add-item` does not appear in the RSF file then it checks to see if the item already exists in the register. If it does, then the RSF file is deemed valid, otherwise it's rejected. This means that an RSF file that has an "orphan entry" could actually be valid if it is a patch on an existing register. 

The inverse is not true for `add-item`. If an `add-item` is in an RSF file, the corresponding `append-entry` must appear in the same RSF file.

content/rsf-spec/index.md
arnau: `alphanum` is defined a few lines below as `alphanum         = ALPHA / DIGIT`, so the `key` definition is: "any ASCII character (A-Z, a-z) or a hypen (%x2D, -) or an underscore (%x5F, _)".

`alpha` and `digit` are defined in https://tools.ietf.org/html/rfc5234#appendix-B.1
`%x2D` and `%x5F` are the hexadecimal codes for hypen and underscore respectively and `/` defines an OR.

content/rsf-spec/index.md
arnau: Good point. Yes `%s` means case sensitive and all tokens that should be strictly in lower case should be prepended with that. I'll amend them 👍

content/rsf-spec/index.md
arnau: Yes, absolutely. I will amend this too 👍

content/rsf-spec/index.md
arnau: I think this RFC should document how things are right now. When we change it, we will have another RFC that explains the change and can refer to the original RFC as its starting point.

content/rsf-spec/index.md
arnau: With the recent changes, is this still the situation? I'll need clarification on this.

content/rsf-spec/index.md
arnau: Ah! I misunderstood this behaviour. I'll amend 👍

content/rsf-spec/index.md
michaelabenyohai: By this I mean that entry numbers are computed independently for `user` and `system` currently (as they are separate logs on the inside).

arnau: Yes, it needs to explain the current implementation exactly.

content/rsf-spec/index.md
michaelabenyohai: Ah I missed the definition of alphanum so that makes sense now. 

Does this imply there is only one character in a key? i.e it's either a single ASCII character, hyphen or underscore. It doesn't appear to say that a key can be made up of multiple of these things but I might be reading it wrong still.

arnau: HA, very good point. I have to add multiple chars :)

I'm wondering if the first char has special treatment in the current implementation? (e.g. can a key start with a hypen: `-foo`? If not, this needs further restriction

content/rsf-spec/index.md
michaelabenyohai: This is currently how it is implemented, yes. ORJ won't let you load RSF that looks like the first example. However, government-service register will currently produce illegal RSF, because we accidentally allowed these duplicate entries into the register before coding the rule.

content/rsf-spec/index.md
arnau: For reference, RFC0003 covers this topic (#16 )

content/rsf-spec/index.md
arnau: Added the description in a new commit.

content/rsf-spec/index.md
arnau: The goal of the spec is to allow different implementations to be built independently and be compatible with each other. If we don't enforce the same rules, a valid RSF in one tool could be invalid in another one. To me, this is a spec failure.

arnau: We can discuss if it should be a rule that exits parsing (and thus rolls back the entire patch), it should be ignored entirely (skipped) or raise a warning but keep parsing.

Given that the current implementation fails, the initial spec should reflect that exact behaviour.

content/rsf-spec/index.md
arnau: You are right, there is a level of checking that guarantees consistency cross log 👍

content/rsf-spec/index.md
arnau: To be clear, `/archive` (zip file) would be normative, rsf a non-normative extension.

MatMoore: That sounds like a good idea - I like the name archive better than register for this purpose, and it makes sense for the zip & rsf to be different representations of the same thing.

PR #13, opened 2018-04-18
Describe more endpoint mappings

arnau:
Signed-off-by: Arnau Siches <arnau.siches@digital.cabinet-office.gov.uk>

### Context

RFC0002 aims to consolidate the API URIs so they can be used more consistently when in CURIE form.

Adds a few more URI mappings for subresources.


arnau:
Note that `/records/{field-name}/{field-value}` is kept as is for now. If we want to make it more distinct we will create a new RFC for it.

PR #21, opened 2018-07-26
Add format reduction RFC

arnau:
Signed-off-by: Arnau Siches <arnau.siches@digital.cabinet-office.gov.uk>

### Context

The Registers data model and its REST API is still in motion. The amount of
formats suggested in the specification and implemented in the reference
implementation increase unnecessarily the complexity of evolving them. For
example, we want to evolve who the schema fits into the data model and how it
is surfaced to the user.

### DIFF
--- a/content/format-reduction/index.md
+++ b/content/format-reduction/index.md
@@ -1,0 +1,52 @@
+---
+rfc:
+start_date: 2018-07-26
+pr:
+status: draft
+---
+
+# Format reduction
+
+## Summary
+
+This RFC proposes to reduce the amount of representations a registers
+implementation should support to JSON and CSV and recommends JSON as the main
+representation. In this way, we expect that implementers, who are not sure of
+which format to implement, will use it by default.
+
+## Motivation
+
+The Registers data model and its REST API are still subject to change. The
+amount of formats which we currently account for unnecessarily complicate the
+implementation. For example, we want to evolve who the schema fits into the
+data model and how the schema is exposed to the user.
+
+
+## Explanation
+
+The current suggestion in the specification is to implement JSON, YAML, CSV,
+TSV, HTML and Turtle. Although none of them are required, the reference
+implementation implements all of them.
+
+Some issues to contextualise this are:
+
+* The specification is not yet mature enough by itself to guide implementers on
+  the implementation of Turtle.
+* The Turtle implementation is incorrect.
+* The YAML representation offers no value on top of the JSON one.
+* The CSV and TSV formats are virtually the same.
+* The HTML representation has diverged from the other representations in the
+  reference implementation, and is no longer an exact analogue. Users have
+  also found the HTML representation confusing in the reference
+  implementation.
+
+As of 2018, users generally favour JSON (89.9%) and a small fraction benefit
+from a tabular form (CSV 3.5%, TSV 1.9%).
+
+In light of that, the suggestion is to remove all formats but JSON and CSV.
+We should bring back RDF when the specification has a mature definition of how to
+implement it. Also, JSON should be recommended as the main representation if
+the implementer has doubts on what format to implement.
+
+Finally, the specification should provide a method for users to ask a register
+for available formats.
--- a/content/format-reduction/index.md
+++ b/content/format-reduction/index.md
@@ -1,8 +1,8 @@
 ---
-rfc:
+rfc: 0004
 start_date: 2018-07-26
-pr:
-status: draft
+pr: openregisters/registers-rfc#21
+status: approved
 ---
 
 # Format reduction
@@ -17,7 +17,7 @@ which format to implement, will use it by default.
 ## Motivation
 
 The Registers data model and its REST API are still subject to change. The
-amount of formats which we currently account for unnecessarily complicate the
+amount of formats which we currently account for unnecessarily complicates the
 implementation. For example, we want to evolve who the schema fits into the
 data model and how the schema is exposed to the user.
 


content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+# Format reduction
+
+## Summary
+
+This RFC proposes to reduce the amount of representations in Registers.
+
+
+## Motivation
+
+The Registers data model and its REST API is still in motion. The amount of
```
bahmady: Rather than 'in motion' I would just emphasise the fact that it's still in alpha, or simply state they are "subject to change".

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+formats suggested in the specification and implemented in the reference
```
bahmady: Could this be rephrased to something like(?):

> The amount of formats which we account for unnecessarily complicate the implementation. 

The current sentence is a little hard to deconstruct.

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+implementation increase unnecessarily the complexity of evolving them. For
+example, we want to evolve who the schema fits into the data model and how it
```
bahmady: Maybe it's just me but I'm not totally clear on how this follows on logically from the previous sentence. In this sentence:

> ... we want to evolve who the schema fits into the data model and how it is surfaced to the user

does _it_ refer to the schema?

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+is surfaced to the user.
+
+
+## Explanation
+
+The current suggestion in the spec is to implement JSON, YAML, CSV, TSV, HTML
```
bahmady: I would say "specification" rather than "spec".

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+and Turtle. Although none of them are required, the reference implementation
+has implemented all of them.
```
bahmady: Perhaps(?):

> ... the reference implementation accounts for all of them

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+There are a few issues to notice:
```
bahmady: Perhaps make it clearer here that this refers to the issues which are the basis for this RFC, i.e. why specific formats should be deprecated

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+The Turtle implementation is not quite
```
bahmady: By "not quite right" is that in the implementation in Registers itself? If so I would point to specific problems with the Turtle implementation itself - at least one.

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+right and the specification is not mature enough to make different
+implementors implement the same thing by following the specification alone.
```
bahmady: Perhaps(?):

> [New sentence]. The specification is not yet mature enough by itself to guide users on the implementation of Turtle.

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+The YAML representation offers no value on top of the JSON one.
+
+The CSV and TSV are virtually the same.
```
bahmady: > CSV and TSV _implementations_

? (or some similar noun)

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+The HTML representtion has been proven confusing for users in the reference
```
bahmady: Typo in 'representation'. Perhaps an alternative paragraph(?):

> The HTML representation has diverged from the other representations in the reference implementation, and is no longer an exact analogue. Users have [also] found the HTML representation confusing in the reference implementation. 

I've put "[also]" because I'm not sure if one follows from the other.

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+implementation and has diverged from being the exact equivalent to the other
+representations so, failing its purpose to be another representation for these
+resources.
+
+The current usage of Registers we can determine that users as per 2018 favour
```
bahmady: > As of DATE-IN-2018, users consume registers in the following formats [list of them with percentages] 

I wouldn't bother repeating the Turtle comment.

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+JSON (89.9%) and a small fraction benefit from a tabular form (CSV 3.5%, TSV
+1.9%). Turtle (1.6%) offers a distinct set of features but as mentioned above
+is not mature enough and it is misleading users.
+
+In light of that, the suggestion is to remove all formats but JSON and CSV.
```
bahmady: I would make this the top of the RFC and then reinforce it here, at the end.

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+start_date: 2018-07-26
+pr:
+status: draft
+---
+
+# Format reduction
+
+## Summary
+
+This RFC proposes to reduce the amount of representations in Registers.
```
MatMoore: This is a minor thing but "in Registers" makes me think less of the presentation formats and more about the internal structure of a register and the data that can be stored in one.

I'd describe this RFC as reducing "the amount of representations a registers implementation should support"

content/format-reduction/index.md
arnau: I like "subject to change"

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+## Explanation
+
+The current suggestion in the spec is to implement JSON, YAML, CSV, TSV, HTML
+and Turtle. Although none of them are required, the reference implementation
+has implemented all of them.
+
+There are a few issues to notice:
+
+The Turtle implementation is not quite
+right and the specification is not mature enough to make different
```
MatMoore: does "right" mean compliant with the spec for Turtle?

content/format-reduction/index.md
MatMoore: there's a typo here: "representtion"

content/format-reduction/index.md
arnau: yes, _it_ refers to the schema.

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+implementors implement the same thing by following the specification alone.
+
+The YAML representation offers no value on top of the JSON one.
+
+The CSV and TSV are virtually the same.
+
+The HTML representtion has been proven confusing for users in the reference
+implementation and has diverged from being the exact equivalent to the other
+representations so, failing its purpose to be another representation for these
```
MatMoore: I can't parse the second half of this sentence - I think it should read "so it is failing its purpose", without the comma.

content/format-reduction/index.md
```diff
@@ -0,0 +1,56 @@
...
+resources.
+
+The current usage of Registers we can determine that users as per 2018 favour
+JSON (89.9%) and a small fraction benefit from a tabular form (CSV 3.5%, TSV
+1.9%). Turtle (1.6%) offers a distinct set of features but as mentioned above
+is not mature enough and it is misleading users.
+
+In light of that, the suggestion is to remove all formats but JSON and CSV.
+And bring back RDF when the specification has a mature definition of how to
```
MatMoore: This sentence doesn't flow naturally to me - can you drop the "and" and change it to "We should bring back RDF when..."

content/format-reduction/index.md
MatMoore: I think this sentence is missing some connecting words. I would just shorten it to something like

"As of 2018, users generally favour JSON (89.9%)..."

content/format-reduction/index.md
arnau: The reference implementation is the one that has an implementation for Turtle that is not quite right.

arnau: ok, "incorrect" it is :)

content/format-reduction/index.md
MatMoore: Are "schema" and "data model" defined terms in the spec right now? If not it may be worth clarifying. It makes sense to me right now, but I'm not sure how much knowledge we should assume in these RFCs.

content/format-reduction/index.md
MatMoore: I would use "representations" for consistency.

content/format-reduction/index.md
arnau: I changed this bit but answering the question: no, the turtle bit (syntax) is fine, the data it expresses is wrong

content/format-reduction/index.md
arnau: Let's see if the new sentence makes more sense

arnau:
@MatMoore @bahmady I accepted most of the suggestions you two made. Could you give it a second read?

Issue #18, opened 2018-06-07
Key (identifier) constrains


arnau:
### Context

The specification doesn't define any restriction for keys. So the only assumption possible is to expect any UTF-8 possible according to JSON rules.

There is an assumption for some components of the team (mainly me) that identifiers are alphanumeric with - or _: `[a-z0-9][a-zA-Z0-9_-]+`. But the reality disagrees.

Also, identifiers are assumed to be case sensitive.

### Explanation

This issue lists and exposes the current registers that use identifiers outside `[a-z0-9][a-zA-Z0-9_-]+`.

* https://www.registers.service.gov.uk/registers/qualification-sector-subject-area — contains `.`
* https://public-body-account-classification.alpha.openregister.org — contains `.`

Identifiers have been provided by custodians (unless they don't have any). That could imply that we should accept anything but knowing that we use CURIEs, and that we want to provide expectations as much as possible, should we set some constrains?

My ideal scenario would've been `[a-z0-9][a-z0-9_]+` but given that reality is more diverse, what else it is safe to add?

@michaelabenyohai @nacnudus thoughts?

/cc @bahmady

nacnudus:
On creation we've always assumed all characters are allowed.  I don't think we've used anything beyond UTF8.   Is there any reason to be more restrictive than URL encoding can handle?

arnau:
URLs have reserved characters (two levels of them) and if these characters are used they need to be percentage-encoded. I would like to avoid as much as possible the need for encoding strategies atop of identifiers.

nacnudus:
What if custodians need them?  The reason for https://github.com/openregister/openregister-java/pull/475 is that forward slashes are part of the existing and widely used IDs of Standard Industrial Classifications.

arnau:
That's why I opened this discussion. I don't think it's acceptable to allow anything and hope for the best. I rather restrict to what we know and assess new things as they happen.

If in the end we accept anything that can be a URL, we need to specify it and make clear if it's a URL specified by WhatWG or by IETF (or if we accept IRIs instead).

Jamie-Atkinson:
I have always assumed that -_,./\ are all allowed within keys and would not blink if someone used one (although I would prefer they did not)

I think there may be a need to allow for circumflexes (â) for Welsh registers etc. But I agree with Arnau that not everything should be included e.g. Œ™© has no place in the identifiers.

nacnudus:
[...]

PR #22, opened 2018-08-02
Add id datatype

arnau:
### Context

Currently, identifiers (i.e. keys) are defined as of any datatype available.
Usage shows only strings have been used thus far.

Allowing any datatype imposes a hard task on the user trying to validate
what to expect in there. Other questions arise when you analyise how each
datatype is expected to behave in all contexts identifiers are used.

This RFC proposes a new datatype to constrain and govern identifiers for
records.

### Guidance to review

Validate the proposal doesn't make any current register invalid, that it is as restrictive as possible but not too restrictive and that all scenarios where an identifier can be used are covered in the proposal.

Also, I would like to know if you think the `/` character can be dropped from the proposal. The only time I've seen it used in identifiers has been in the work done for SIC codes https://github.com/openregister/industrial-classification-data but this is not yet a register.

Also, in #18 we discussed about alphabetic characters outside the ASCII range (e.g. welsh). If you think it's important to include them let me know. Keep in mind that the consequence of that would be having CURIEs with percent-encoded characters as CURIEs need to play nice with https://url.spec.whatwg.org/

Now, if it's imperative to find a balance an option to consider is to change the definition of CURIE to something different that only requires percent-encoding once expanding to a URL. It's a trade-off we could consider making.


content/id-datatype/index.md
```diff
@@ -0,0 +1,91 @@
...
+currently used:
+
+```abnf
+id = (ALPHA / DIGIT) *(recommended / legacy)
+recommended = ALPHA / DIGIT / restricted
+restricted = "-" / "_" / "."
+legacy = "%2F"
+```
+
+Where `restricted` and `legacy` can't appear consecutively.
```
gidsg: I found this wording ambiguous. I think you mean any characters in the groups 'restricted' or 'legacy' cannot appear consecutively in a valid key?
The current wording might be taken to mean specifically legacy characters cannot appear directly after restricted ones, which is not the case (at least that I can tell from the examples.)

content/id-datatype/index.md
```diff
@@ -0,0 +1,94 @@
...
+means that a CURIE must always percent-encode this character as `%2F`.
+
+The proposal then is to not allow more characters than the ones that are
+currently used:
+
+```abnf
+id = (ALPHA / DIGIT) *(ALPHA / DIGIT / restricted)
+restricted = delimiters / legacy
+delimiters = "-" / "_" / "."
+legacy = "%2F"
```
MatMoore: This implies '/' is always percent encoded in IDs but the text above says it's only percent encoded when used as a CURIE.

arnau: Very true, I'll amend this, thanks

arnau:
> So I'm wondering if we could actually go one step further and define a preferred way of delimitating segments of an ID when there is no preexisting standard. For example, recommending hyphens over underscores, periods or slashes.

At the spec level I'm happy to add an informative block suggesting the preferred delimiter. In practice, our current conventions say that if a custodian already has identifiers, we should use them. I'm split on this because although I agree with you on odd looking "200010.11" IDs, I see the value in having familiar identifiers such as "GB" (country register).

MatMoore:
I think @Jamie-Atkinson and @trvllngmn should be the ones deciding this, but IMO the consumer's familiarity with the existing IDs is way more important than what the custodian is already doing. I think "GB" is a great id because it's already a widely used standard and it's human readable.

nacnudus:
IMO "GB" is a bad ID because it's a standard that the custodian can't control, is human readable (meaningful), and can be reassigned.

> After a country is deleted from ISO 3166-1, its alpha-2 and alpha-3 codes will be transitionally reserved for a transitional period of at least fifty years. After the expiration of the transitional period, these codes are free to be reassigned. [Wikipedia](https://en.wikipedia.org/wiki/ISO_3166-3)

Registers IDs should be meaningless by definition, and then there will be no confusion about decimal points or whatever.  "Any resemblance to IDs living or dead is purely coincidental".  Then we can prioritise IDs that integrate easily with custodians' existing systems.

content/id-datatype/index.md
```diff
@@ -0,0 +1,103 @@
...
+Evidence says that identifiers are composed primarily by alphanumeric ASCII
+characters with a few other printable characters:
+
+* `-` (U+002D)
+* `_` (U+005F)
+* `.` (U+002E)
+* `/` (U+002F)
+
+Note that `/` has meaning in URLs and hence it has meaning in CURIEs. This
+means that a CURIE must always percent-encode this character as `%2F`.
```
michaelabenyohai: Does this mean that a valid curie would be "country:a%2Fb" and it would resolve to https://country.register.gov.uk/records/a%2Fb? Whereas an invalid curie would be "country:a/b" as this would resolve to https://country.register.gov.uk/records/a/b? 

My question is whether we could allow "/" in curies (such that "country:a/b" is valid) but dictate that it is the responsibility of those interpreting a curie to resolve it to https://country.register.gov.uk/records/a%2Fb? However, I guess this would restrict us from being able to interpret "country:a/b" as https://country.register.gov.uk/records/a/b in the future if we wanted to. So I think it makes sense to either encode it in the curie or drop them altogether for now.

Given that the industrial-classification registers are still in alpha, we will need to either re-load them to encode the `/` characters in their IDs or we could use new keys. I'm not sure what the status of these registers are.

arnau:
> My only question is what we mean by ID datatype.
>
> Is this a datatype in the sense that it belongs alongside the other datatypes in the datatype register? > And that a field in an item can have this datatype?

To summarise our conversation outside this thread: This RFC has two overlapping intentions and it shouldn't. 1) Aims to restrict the values an entry key can have; 2) Aims to provide a datatype that can be used in data.

I'll revisit the RFC and focus on the entry key restriction, which means it doesn't need to be a datatype.

The restricted datatype will be addressed in a different RFC with a name that doesn't suggest unique identification to better split concepts.

content/id-datatype/index.md
arnau: > Does this mean that a valid curie would be "country:a%2Fb" and it would resolve to https://country.register.gov.uk/records/a%2Fb? Whereas an invalid curie would be "country:a/b" as this would resolve to https://country.register.gov.uk/records/a/b?

Correct.

> My question is whether we could allow "/" in curies (such that "country:a/b" is valid) but dictate that it is the responsibility of those interpreting a curie to resolve it to https://country.register.gov.uk/records/a%2Fb? However, I guess this would restrict us from being able to interpret "country:a/b" as https://country.register.gov.uk/records/a/b in the future if we wanted to. So I think it makes sense to either encode it in the curie or drop them altogether for now.

Yes, if we diverge from the standard definition of how a CURIE works, it means we will have to define the rules to expand to something like https://country.register.gov.uk/records/a/b

I rather stick to CURIEs and try to minimise the use of `/` in identifiers.

> Given that the industrial-classification registers are still in alpha, we will need to either re-load them to encode the / characters in their IDs or we could use new keys. I'm not sure what the status of these registers are.

Umm, I need to address the ID definition because as IDs it is valid to use "a/b" but when they are part of a CURIE or a URL, they need to be encoded.

arnau:
@MatMoore @michaelabenyohai I changed the wording referring to "id datatype" to re-scope the RFC to entry keys as they are the real identifiers.

Also, I tried to make a bit more clear that an identifier such as "a/b" has to be stored this way and only percent encoded when in a CURIE or URL.

PR #23, opened 2018-08-02
RFC0009: Entry hash

arnau:
### Context

The current specfication doesn't define what is the algorithm to get from the
entry data to the entry hash for the given hashing algorithm (from now on
SHA-256). The reference implementation (ORJ) reuses the item canonicalisation
algorithm to canonicalise the entry JSON representation and then hash the
result. Finally, relying on the JSON representation has the downside of being
more rigid than it could be.

This RFC proposes an algorithm to hash entries without relying in JSON or in
the attribute names.

**This RFC is incompatible with the reference implementation**

### Guidance to review

Review it makes sense and it has everything needed to be implemented.

### DIFF
--- a/content/entry-hash/index.md
+++ b/content/entry-hash/index.md
@@ -69,9 +69,10 @@ as the [objecthash](https://github.com/benlaurie/objecthash) algorithm.
 Tags:
 
 * `i`: Integer.
-* `s`: String (e.g. `ID` type).
+* `u`: Unicode String (e.g. `ID` type).
 * `t`: Timestamp.
-* `h`: Hash.
+* `r`: Raw Hash.
+* `l`: List.
 
 ```elm
 hashValue : Alg -> Value -> Tag -> Hash
@@ -92,7 +93,7 @@ timestampHash = hashValue Sha256 "2016-04-05T13:23:05Z" Timestamp
 timestampHash == "f22ecc4464f22c8fee624769189665a0afd7ef10a2775a000082c47cbd9f6419"
 
 keyHash = hashValue Sha256 "GB" String
-keyHash == "c8fe89332c2aa2b92234bd43d0fb327c64fdbdce4ea527f51f78111543dd5341"
+keyHash == "fff7021c7df4426be0f9a3c83f236eb6f85d159e624b010d65e6dde267889c21"
 ```
 
 The item is a special case:
@@ -104,14 +105,48 @@ The item is a special case:
 The algorithm:
 
 1. Foreach value, strip out the prefix.
-2. Concatenate all hashes in strict order.
-3. Hash the result.
+2. Tag and Hash each value.
+3. Concatenate all hashes in the same order.
+4. Tag and hash the result.
 
 ```elm
 hashValueList : Alg -> List Value -> Tag -> Hash
 
 itemHash  = hashValueList Sha256 ["sha-256:6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"] Hash
-itemHash == "2d44caa5719d20a648aff8de575aa4832209ae0274f66edf5c33bcebceefef9a"
+itemHash == "06e387ba89400ba977e5a146c567cf7337db4de1dd41498366e30d6e77b5b2bc"
+```
+
+These are the steps for the example above:
+
+1. Strip out prefixes:
+
+```elm
+rawList = stripPrefix ["sha-256:6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"]
+rawList == ["6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"]
+```
+
+2. Tag and hash:
+
+```elm
+hashedList =
+  map hashListValue ["6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"]
+  where
+    hashListValue e = hashValue Sha256 e Hash
+
+hashedList == ["63825e4c6f3dd8b229be451b17f57a4e431eb92ae62f9581c9bd268558c27177"]
+```
+3. Concatenate:
+
+```elm
+concatHash = concatAll ["63825e4c6f3dd8b229be451b17f57a4e431eb92ae62f9581c9bd268558c27177"]
+concatHash == "63825e4c6f3dd8b229be451b17f57a4e431eb92ae62f9581c9bd268558c27177"
+```
+
+4. Tag and hash:
+
+```
+itemHash = hashValue Sha256 concatHash List
+itemHash == "06e387ba89400ba977e5a146c567cf7337db4de1dd41498366e30d6e77b5b2bc"
 ```
 
 ### order the resulting hashes
--- a/content/entry-hash/index.md
+++ b/content/entry-hash/index.md
@@ -1,8 +1,9 @@
 ---
 rfc: 0009
 start_date: 2018-08-02
+decision_date: 2018-08-30
 pr: openregister/registers-rfcs#23
-status: draft
+status: approved
 ---
 
 # Entry hash
--- a/content/entry-hash/index.md
+++ b/content/entry-hash/index.md
@@ -1,0 +1,141 @@
+---
+rfc:
+start_date: 2018-08-02
+pr:
+status: draft
+---
+
+# Entry hash
+
+## Summary
+
+This RFC proposes an algorithm to hash entries without relying in JSON or in
+the attribute names.
+
+**This RFC is incompatible with the reference implementation**
+
+## Motivation
+
+The current specfication doesn't define what is the algorithm to get from the
+entry data to the entry hash for the given hashing algorithm (from now on
+SHA-256). The reference implementation (ORJ) reuses the item canonicalisation
+algorithm to canonicalise the entry JSON representation and then hash the
+result. Finally, relying on the JSON representation has the downside of being
+more rigid than it could be.
+
+## Explanation
+
+The algorithm has three parts:
+
+1. tag and hash each value in the entry.
+2. order the resulting hashes.
+3. hash the result of concatenating all hashes.
+
+For reference, this is the abstract definition of an entry:
+
+```elm
+type Entry =
+  { number: Integer
+  , key: ID
+  , timestamp: Timestamp
+  , item: List Hash
+  }
+```
+
+And this is the current serialisation in JSON:
+
+```json
+[
+  {
+    "index-entry-number":"6",
+    "entry-number":"6",
+    "entry-timestamp":"2016-04-05T13:23:05Z",
+    "key":"GB",
+    "item-hash":["sha-256:6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"]
+  }
+]
+```
+
+Notice they are not exaclty the same even thought they represent the same
+entity. Also notice that the `ID` type is not yet an approved RFC. For this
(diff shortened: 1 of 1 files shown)


content/entry-hash/index.md
```diff
@@ -0,0 +1,176 @@
...
+pr:
+status: draft
+---
+
+# Entry hash
+
+## Summary
+
+This RFC proposes an algorithm to hash entries without relying in JSON or in
+the attribute names.
```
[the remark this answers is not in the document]
arnau: Fair question; the problem is that the data in the entry is mixed with the serialisation of the structure of the entry and, because the hash is for integrity purposes, if I decide to change how I represent the entry, but not the entry itself, I'm changing its identity. I should explain this in the RFC 👍

content/entry-hash/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+**This RFC is incompatible with the reference implementation**
```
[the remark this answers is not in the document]
arnau: It is a disadvantage indeed. The problem is that currently the hashing for entries relies on the autogenerated JSON out of the entry entity which happens to include the index entry number which may or may not be part of the serialisation but it surely is not part of the entry.

content/entry-hash/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Motivation
+
+The current specfication doesn't define what is the algorithm to get from the
+entry data to the entry hash for the given hashing algorithm (from now on
+SHA-256). The reference implementation (ORJ) reuses the item canonicalisation
+algorithm to canonicalise the entry JSON representation and then hash the
+result. Finally, relying on the JSON representation has the downside of being
```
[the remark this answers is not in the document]
arnau: 👍 I'll explain that along the previous comment you made.

content/entry-hash/index.md
```diff
@@ -0,0 +1,176 @@
...
+more rigid than it could be.
+
+## Explanation
+
+The algorithm has three parts:
+
+1. tag and hash each value in the entry.
+2. order the resulting hashes.
+3. hash the result of concatenating all hashes.
```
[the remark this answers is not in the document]
arnau: umm no, this is the algorithm for the entry hash, the item hash is different. I'm working on another RFC for that one due the redactable stream of work but they are different.

The entry has a fixed structure and it is known by the register, the item has a flexible structure that changes from register to register and when we have schema evolution it will change over time as well.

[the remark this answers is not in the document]
arnau: In other words, if the item changes, its hash changes and, because the entry depends on the item hash to generate its own hash, it will change as well.

arnau:
FYI, if it helps in any way, this is the implementation I used to describe the step by step section https://github.com/arnau/elm-objecthash/blob/sketch/tests/SketchTest.elm

content/entry-hash/index.md
```diff
@@ -0,0 +1,219 @@
...
+
+```elm
+keyValue = toString (ID "GB")
+keyValue == "GB"
+
+keyHash = hashValue Sha256 keyValue String
+toHex keyHash == "fff7021c7df4426be0f9a3c83f236eb6f85d159e624b010d65e6dde267889c21"
+```
+
+The item is a special case:
```
MatMoore: Does this need to be a special case? Can't we just treat it as a string and follow the normal objecthash algorithm? If so that would simplify the implementation a lot as you just need to implement the objecthash algorithm once.

arnau: Well, objecthash accounts for "raw" byte sequences so from that perspective it is not a special case. It is though with the rest of values in Entry. The wording is not clear enough.

arnau:
> This makes sense now that I've read the item hash RFC. However, I would want to use the same algorithm for both hashes if possible.

Well, the algorithm is quite similar, the only manual step is the composition of the initial list of values (as opposed to use a dictionary). That's is required if I want to avoid relying on specific attribute names.

content/entry-hash/index.md
```diff
@@ -0,0 +1,219 @@
...
+
+It is a set of values and each value is already hashed. This means that the
+content of this set only needs a sorted concatenation. Then, tag it as a set
+`s` and hash the result.
+
+```elm
+hashValueSet : Alg -> Set Hash -> Tag -> Hash
+
+itemHash = hashValueSet Sha256 (Set [Sha256 "6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"]) Hash
+toHex itemHash == "94b34bd5705b1a8c31796c00347dad2f5a1364050a1f6b768c7ac289502f62ce"
```
michaelabenyohai: Can you clarify how you got this result? I have tried working through the example and can't get the same result.

content/entry-hash/index.md
```diff
@@ -0,0 +1,219 @@
...
+
+### hash the result of concatenating all hashes
+
+And finally, tag and hash the result of concatenating the list of hashes:
+
+```elm
+hash : Alg -> List Hash -> Hash
+
+entryHash = hash Sha256 hashList
+entryHash == "cc33f805025fb5bcd43cd9a1d0d77093af443106ff1513a2f5dfff2b81ae4fd3"
```
michaelabenyohai: I also can't get this result from trying the example? Can you clarify, perhaps in code, how you did it?

content/entry-hash/index.md
```diff
@@ -0,0 +1,219 @@
...
+5. Let _hashList_ be an empty list.
+6. Apply the _hashValue_ function to _number_ tagged with `i` (Integer). And
+   append the result to _hashList_.
+7. Apply the _hashValue_ function to _key_ tagged with `u` (String). And
+   append the result to _hashList_.
+8. Apply the _hashValue_ function to _timestamp_ tagged with `t` (Timestamp). And
+   append the result to _hashList_.
+9. Map over each value _item_ in the _items_ set:
+   1. Apply the _hashValue_ function to _item_ tagged with `r` (Hash).
+10. Sort the elements of _items_, concatenate them and
```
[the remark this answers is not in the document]
arnau: I'll make it more clear. I need to review this because the treatment for raw values is not clear enough.

content/entry-hash/index.md
```diff
@@ -0,0 +1,219 @@
...
+When this algorithm operates on _hashes_ (e.g. tag, concatenate) it is done on
+the byte level, not the hexadecimal string representation that the latter
+example shows as partial representations.
+
+1. Let _number_ be the string representation of the entry number.
+2. Let _key_ be the string representation of the entry key.
+3. Let _timestamp_ be the string representation of the entry timestamp.
+4. Let _items_ be the set of item hashes as bytes. Note that
+   the item hashes must not have the prefix (i.e. `sha-256:`) when transformed to
+   bytes.
```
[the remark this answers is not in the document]
arnau: Mostly because the hash is treated as a byte sequence and the `sha-256` is not part of the hash. When treated like a byte sequence, you don't need to hash it again (raw), you can use it as is, which is similar to how redaction works.

I've always seen the hashes with the prefix as a string representation of a qualified sequence of bytes encoded in hexadecimal. That's probably the reason.

content/entry-hash/index.md
michaelabenyohai: I've got the same result as you now and realise where the confusion lies.

In the step by step description you say this:

For each value item in the items set:
- Let _items_ be the set of item hashes as bytes. Note that
   the item hashes must not have the prefix (i.e. `sha-256:`) when transformed to
   bytes.
- Let _hashList_ be an empty list.
- Let _itemsBytes_ be
- For each value _item_ in the _items_ set:
- Apply the hashValue function to item **tagged with r (Hash)** and append the result to itemsBytes.
- Sort the elements of itemsBytes, concatenate them and
- Apply the hashValue function tagged with s (Set). And append the result to hashList.

From this, I did the following for the example:
1. Get the bytes for "6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"
2. Prepend the result of 1 with "r"
3. Hash the result of 2
4. Get the bytes of the result from 3 and prepend it with "s"
5. Hash the result of 4

And I got the result "cff910f74878650a3cceb54039bdb62707de9d20e80d4385127732a4e444bd57".

If I instead do:
1. Get the bytes for "6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"
2. Get the result from 1 and prepend it with "s"
3. Hash the result of 2

And I get your result of "94b34bd5705b1a8c31796c00347dad2f5a1364050a1f6b768c7ac289502f62ce".

To get your answer it appears to miss out the below step from your description:

- Apply the hashValue function to item **tagged with r (Hash)** and append the result to itemsBytes.

Does this need to be removed?

content/entry-hash/index.md
michaelabenyohai: I managed to get this. I was just doing it wrong. So ignore this comment!

content/entry-hash/index.md
arnau: The error was in the example, I wasn't tagging the raw values with `r`. I reviewed the original objecthash implementation and the algorithm is correct.

So, the values you were getting were right. I amended and hopefully clarified the examples a bit more.

content/entry-hash/index.md
```diff
@@ -0,0 +1,239 @@
...
+
+The sorting algorithm for a set of hashes is done by comparing the list of
+bytes one by one. For example, given a set `["foo", "bar"]` you'll get the
+folllowing byte lists after hashing them as unicode:
+
+```elm
+[ [166,166,229,231,131,195,99,205,149,105,62,193,137,194,104,35,21,217,86,134,147,151,115,134,121,181,99,5,242,9,80,56]
+, [227,3,206,11,208,244,193,253,254,76,193,232,55,215,57,18,65,226,224,71,223,16,250,97,1,115,61,193,32,103,93,254]
+]
+```
```
michaelabenyohai: To get this result you must prepend "foo" and "bar" with 'u'. This sorting algorithm is used in step 11 where the individual components of _itemBytes_ have been preprended with 'r'. Is it worth changing this example to make it clear that 'r' is used to tag and not 'u'?

michaelabenyohai: Or just make it clear that in this example the strings have been tagged with `u` though the bytes may have come from any tag and any hash.

michaelabenyohai: I think I might have changed my mind about this, it does specify "hashing them as unicode" which implied the tag 'u' is used. Ignore me.

content/entry-hash/index.md
```diff
@@ -0,0 +1,239 @@
...
+   the item hashes must not have the prefix (i.e. `sha-256:`) when transformed to
+   bytes.
+5. Let _hashList_ be an empty list.
+6. Apply the _hashValue_ function to _number_ tagged with `i` (Integer). And
+   append the result to _hashList_.
+7. Apply the _hashValue_ function to _key_ tagged with `u` (String). And
+   append the result to _hashList_.
+8. Apply the _hashValue_ function to _timestamp_ tagged with `t` (Timestamp). And
+   append the result to _hashList_.
+9. Let _itemsBytes_ be
```
MatMoore: is part of this step missing?

content/entry-hash/index.md
```diff
@@ -0,0 +1,239 @@
...
+    "index-entry-number":"6",
+    "entry-number":"6",
+    "entry-timestamp":"2016-04-05T13:23:05Z",
+    "key":"GB",
+    "item-hash":["sha-256:6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"]
+  }
+]
+```
+
+Notice they are not exaclty the same even thought they represent the same
```
MatMoore: there's a typo here: "thought"

PR #24, opened 2018-08-06
RFC0010: Item hash

arnau:
### Context

The current item hashing algorithm makes impossible to redact a bit of data inside the item. All you can do is to redact the full blob. This don't end well when for example a GDPR request forces redacting a piece of information replicated in many items. Which is likely to happen given that every time a new item is introduced with a change, the rest of the data is copied.

This RFC proposes an algorithm to hash items such that data can be redacted without breaking the integrity of the register.

This RFC is not backwards compatible

### Guidance to review

Ensure the algorithm is unequivocally described such that two independent implementations would result in the same behaviour.

### DIFF
--- a/content/item-hash/index.md
+++ b/content/item-hash/index.md
@@ -10,7 +10,7 @@ status: draft
 ## Summary
 
 This RFC proposes an algorithm to hash items such that data can be redacted
-without breaking the integrity of the register.
+at any granularity without breaking the integrity of the register.
 
 **This RFC is not backwards compatible**
 
--- a/content/item-hash/index.md
+++ b/content/item-hash/index.md
@@ -87,7 +87,7 @@ intended to be forward compatible.
 
 ### String normalisation algorithm
 
-1. For ASCII control characters (codepoints 0x00 - 0x1f):
+1. For control characters (codepoints 0x00 - 0x1f):
    1. If it has a short representation (`\b`, `\f`, `\n`, `\r`, or `\t`), that
       short representation MUST be used.
    2. Other control characters (such as `NULL`) MUST be represented as a
@@ -117,25 +117,15 @@ i_1 =
        ]
 ```
 
-Which in JSON are represented as:
-
-```json
-{
-  "foo": "abc",
-  "bar": "xyz"
-}
-```
-
-And
-
-```json
-{
-  "foo": "**REDACTED**2a42a9c91b74c0032f6b8000a2c9c5bcca5bb298f004e8eff533811004dea511",
-  "bar": "xyz"
-}
-```
+The reason they are equivalent is because the hash for `"abc"` is
+"2a42a9c91b74c0032f6b8000a2c9c5bcca5bb298f004e8eff533811004dea511". So when
+the hashing algorithm is applied to the whole item, the redacted value is used
+as is (without the redaction tag).
 
 ### Example
+
+Note: To simplify the code, the hashing algorithm `SHA-256` is implied. A full
+implementation must acknowledge the possibility of a different algorithm.
 
 To walk through the algorithm I'll use the following item:
 
@@ -144,7 +134,7 @@ Dict
   [ ("id", "GB")
   , ("official-name", "The United Kingdom of Great Britain and Northern Ireland")
   , ("name", "United Kingdom")
-  , ("citizen-names", ["Briton", "British citizen"])
+  , ("citizen-names", Set ["Briton", "British citizen"])
   ]
 ```
 
@@ -157,4 +147,89 @@ Which is serialised as JSON as:
   "name": "United Kingdom",
   "citizen-names": ["Briton", "British citizen"]
 }
+```
+
+The following definitions set the context of operation:
+
+```elm
+type Tag
+  = Dict
+  | Set
+  | String
+
+tagToChar : Tag -> Char
+tagToChar tag =
+  case tag of
+    Dict ->
+      'd'
+    Set ->
+      's'
+    String ->
+      'u'
+```
+
+```elm
+hash : Item -> Hash
+```
+
+```elm
+hashString : String -> Hash
+
+hashSet : Set -> Hash
+
+hashValue : Value -> Hash
+hashValue value =
+  case value of
+    VString str ->
+      if startsWith "**REDACTED**" str then
+        hashString (dropLeft 12 str)
+      else
+        hashString str
+    VSet set ->
+      hashSet set
+```
+
+The `hash` function, step by step would roughly look like:
+
+```elm
+id =
+  concat ["17b788a70eeccbdc2fcb2d2d3db216c02fa88ac668beeb164bb2328c864bf3f4", "fff7021c7df4426be0f9a3c83f236eb6f85d159e624b010d65e6dde267889c21"]
+
+id == "d62311d17806761e4195bda80a4ba7f20c8cacc58c3f6f0f1a22e64f47a77651"
+
+officialName =
+  concat ["cf09bea8c0107bd2150b073150d48db0a5b24c83defc7960ed698378d9f84b93", "bf1860175c77869938cf9f4b37edb00f2f387be7b361f9c2c4a2ac202c1ba2e5"]
+
+officialName == "103ea25f856f755082745f183b1f6f41c08b1c35df794d1382650637f35d0949"
+
+name =
+  concat ["5c0be87ed7434d69005f8bbd84cad8ae6abfd49121b4aaeeb4c1f4a2e2987711", "94099b1e0b9a1e673bafee513080197fa1980895ca27e091fdd4c54fab2bed24"]
+
+name == "df50f7bd9e6af8c5dd009f99e66a0192ca6204c4b7e45ee79e4a89d1d7ace200"
+
+citizenNameSet =
+  ["3d76c67f95cb9c4fc8e9dfdaa1d0ac4cbf6feba4dc7521429618afad925a3922", "f59ca1c4c086398b98aa6ad3ab2b14c4c1a15c225da73c706f5023115a7acd50"]
+  |> concat
+  |> cons 's'
+  |> hashTagged
+
+citizenNameSet == "1b68822ac12017ae10eebcce34c4cd5e07d83b6c76bdca8f14eb54ab60096269"
+
+citizenNames =
+  concat ["bb3a7ac86d4f90c20d099992de0bd09bf3c4f27169c2cd873836762b01d5a2be", "1b68822ac12017ae10eebcce34c4cd5e07d83b6c76bdca8f14eb54ab60096269"]
+
+citizenNames == "c8a2625212e2bac74346c60ae4988aa9978958918bd1455e35e6d0670af15ff3"
+```
+
+Then sort and aggregate the partial results:
+
+```elm
+itemHash =
+  [id, officialName, name, citizenNames]
+  |> sort
+  |> concat
+  |> cons 'd'
+  |> hashTagged
+
+itemHash == "5bc0163d594fb6e958d2758eff074fb4d25cd3f3867ff30e9cbe982c59cb90b5"
 ```
--- a/content/item-hash/index.md
+++ b/content/item-hash/index.md
@@ -1,0 +1,160 @@
+---
+rfc:
+start_date: 2018-08-03
+pr:
+status: draft
+---
+
+# Item hash revisit
+
+## Summary
+
+This RFC proposes an algorithm to hash items such that data can be redacted
+without breaking the integrity of the register.
+
+**This RFC is not backwards compatible**
+
+
+## Motivation
+
+The current item hashing algorithm makes impossible to redact a bit of data
+inside the item. All you can do is to redact the full blob. This don't end
+well when for example a GDPR request forces redacting a piece of information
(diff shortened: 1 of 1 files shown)


arnau:
FYI, if it helps in any way, this is the implementation I used to describe the step by step section https://github.com/arnau/elm-objecthash/blob/sketch/tests/SketchTest.elm

content/item-hash/index.md
```diff
@@ -0,0 +1,213 @@
...
+start_date: 2018-08-03
+pr:
+status: draft
+---
+
+# Item hash with redaction
+
+## Summary
+
+This RFC proposes an algorithm to hash items such that data can be redacted
```
MatMoore: I guess this should say data can be partially redacted, since we could redact now if we wanted to redact the whole item.

content/item-hash/index.md
```diff
@@ -0,0 +1,213 @@
...
+without breaking the integrity of the register.
+
+**This RFC is not backwards compatible**
+
```
MatMoore: Are there any security considerations for this RFC?

arnau:
> I'd like to read through all the related RFCs and try to understand better how future registers will work with all these changes, so I may come up with more comments or questions later.

Sure!

content/item-hash/index.md
```diff
@@ -0,0 +1,213 @@
...
+The current item hashing algorithm makes impossible to redact a bit of data
+inside the item. All you can do is to redact the full blob. This don't end
+well when for example a GDPR request forces redacting a piece of information
+replicated in many items. Which is likely to happen given that every time a
+new item is introduced with a change, the rest of the data is copied.
+
+## Explanation
+
+The following explanation is an implementation of the [objecthash](https://github.com/benlaurie/objecthash)
```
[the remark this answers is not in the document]
arnau: Correct.

content/item-hash/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+### String normalisation algorithm
+
+1. For control characters (codepoints 0x00 - 0x1f):
+   1. If it has a short representation (`\b`, `\f`, `\n`, `\r`, or `\t`), that
+      short representation MUST be used.
+   2. Other control characters (such as `NULL`) MUST be represented as a
+      `\u00XX` escape sequence. Hexadecimal digits MUST be upper-case.
+2. Backslash (`\`) and double quote (`"`) MUST be escaped as `\\` and `\"`
```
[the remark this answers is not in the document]
arnau: My thinking is that to reduce the chances of discrepancies in hashing, types with arbitrary unicode codepoints (i.e. strings) have a canonical form. That said, I should verify this assumption and see if I'm trying to solve a problem that doesn't really exist.

content/item-hash/index.md
arnau: What kind of security considerations are you thinking on?

MatMoore: I don't have any specific concerns, it just seems like a section we should include in any RFC that touches on integrity guarantees or anything like that (I also noticed in the current spec, the security considerations section is blank). As a reader I don't know what I don't know, so even if there's nothing to say here it's good to have something that says so.

MatMoore:
I just wanted to say I looked at the elm code and that cleared up a lot of my questions. Sorry for the barrage of comments yesterday!

For some reason I misread one of the steps and thought the keys were not part of the hash, but they are included when computing the item hash, so both algorithms make sense to me now.

My only outstanding concern with these RFCs is whether we need the string normalisation part or whether we can just hash the string as is.

content/item-hash/index.md
```diff
@@ -0,0 +1,213 @@
...
+       ]
+
+i_1 =
+  Dict [ ("foo", "**REDACTED**2a42a9c91b74c0032f6b8000a2c9c5bcca5bb298f004e8eff533811004dea511")
+       , ("bar", "xyz")
+       ]
+```
+
+The reason they are equivalent is because the hash for `"abc"` is
+"2a42a9c91b74c0032f6b8000a2c9c5bcca5bb298f004e8eff533811004dea511". So when
```
gidsg: maybe I am misunderstanding, but if I do a sha-256 hash of `abc` I get a different value: `ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad`?

content/item-hash/index.md
```diff
@@ -0,0 +1,213 @@
...
+citizenNames == "bb3a7ac86d4f90c20d099992de0bd09bf3c4f27169c2cd873836762b01d5a2be16897987a6ee59d9ffdb456ed02df34a79b05346498d4360172568101ae157c1"
+```
+
+Then combine the partial results and tag it as a dictionary:
+
+```elm
+itemHash = hashDict [id, officialName, name, citizenNames]
+
+itemHash == "45d9392ad17cead3fa46501eba3e5ac237cb46a39f1e175905f00ef6a6667257"
+```
```
gidsg: maybe would be worth adding an example where a value is subsequently redacted?

content/item-hash/index.md
gidsg: or do you mean the hash of the value of `abc`?

content/item-hash/index.md
```diff
@@ -0,0 +1,213 @@
...
+            tag it with `u` (String), hash it and append it to _elList_.
+         3. Concatenate _elList_ elements, tag it with `s` (Set), hash it and set
+            it to _valueHash_.
+   3. If _value_ starts with `**REDACTED**`, set _normValue_ with _value_
+      without `**REDACTED**`.
+   4. Otherwise, normalise _value_ according to [string normalisation](#string-normalisation-algorithm)
+      tag it with `u` (String), hash it and set _valueHash_.
+   5. Tag _attr_ with `u` (String), hash it and set _attrHash_.
+   6. Concat _attrHash_ and _valueHash_ and append to _hashList_.
+3. Sort _hashList_.
```
gidsg: Is the sorting algorithm defined?

content/item-hash/index.md
gidsg: OK, it's because I was missing the tag prefix for string `u`.

content/item-hash/index.md
```diff
@@ -0,0 +1,213 @@
...
+2. Let _hashList_ be an empty list.
+2. Foreach _(attr, value)_ pair in _item_:
+   1. If _value_ is null, continue.
+   2. If _value_ is a Set:
+      1. Let _elList_ be an empty list.
+      2. Foreach _el_ in _value_:
+         1. If _el_ starts with `**REDACTED**`, append _el_ without `**REDACTED**`
+            to _elList_.
+         2. Otherwise, normalise _el_ according to [string normalisation](#string-normalisation-algorithm)
+            tag it with `u` (String), hash it and append it to _elList_.
```
gidsg: can we make it more explicit that the tag is prepended?

content/item-hash/index.md
arnau: I'll add it 👍

content/item-hash/index.md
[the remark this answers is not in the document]
arnau: Thinking about this, perhaps the only normalisation required is ensuring there are no composite characters (NFC vs NFD).

https://en.wikipedia.org/wiki/Unicode_equivalence

arnau:
@MatMoore @gidsg I tried to address all your comments, let me know if it is satisfactory.

content/item-hash/index.md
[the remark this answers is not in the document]
arnau: I'm convinced now, NFC is the only normalisation we need here 👍 

Most languages either directly represent strings in NFC or the have ways in stdlib to do so.

content/item-hash/index.md
```diff
@@ -0,0 +1,257 @@
...
+citizenNames = hashPair ("citizen-names", Set ["Briton", "British citizen"])
+citizenNames == "bb3a7ac86d4f90c20d099992de0bd09bf3c4f27169c2cd873836762b01d5a2be16897987a6ee59d9ffdb456ed02df34a79b05346498d4360172568101ae157c1"
+```
+
+Then combine the partial results and tag it as a dictionary:
+
+```elm
+itemHash = hashDict [id, officialName, name, citizenNames]
+
+itemHash == "45d9392ad17cead3fa46501eba3e5ac237cb46a39f1e175905f00ef6a6667257"
```
michaelabenyohai: I can't get this value. It would be good to show you what I'm doing to see where I'm going wrong.

content/item-hash/index.md
```diff
@@ -0,0 +1,257 @@
...
+  , ("citizen-names", Set ["Briton", "British citizen"])
+  ]
+```
+
+To:
+
+```elm
+Dict
+  [ ("id", "GB")
+  , ("official-name", "**REDACTED**cf09bea8c0107bd2150b073150d48db0a5b24c83defc7960ed698378d9f84b93bf1860175c77869938cf9f4b37edb00f2f387be7b361f9c2c4a2ac202c1ba2e5")
```
michaelabenyohai: Should this instead be:

```
("official-name","**REDACTED**bf1860175c77869938cf9f4b37edb00f2f387be7b361f9c2c4a2ac202c1ba2e5")
```

content/item-hash/index.md
```diff
@@ -0,0 +1,257 @@
...
+   2. If _value_ is a Set:
+      1. Let _elList_ be an empty list.
+      2. Foreach _el_ in _value_:
+         1. If _el_ starts with `**REDACTED**`, append _el_ without `**REDACTED**`
+            to _elList_.
+         2. Otherwise, normalise _el_ according to [string normalisation](#string-normalisation-algorithm)
+            tag it with `u` (String), hash it and append it to _elList_.
+      3. Concatenate _elList_ elements, tag it with `s` (Set), hash it and set
+         it to _valueHash_.
+   3. If _value_ starts with `**REDACTED**`, set _normValue_ with _value_
```
michaelabenyohai: Should _normValue_ be _valueHash_?

content/item-hash/index.md
michaelabenyohai: I got this after trying again so ignore this comment!

content/item-hash/index.md
```diff
@@ -0,0 +1,257 @@
...
+### Algorithm
+
+When this algorithm operates on _hashes_ (e.g. tag, concatenate) it is done on
+the byte level, not the hexadecimal string representation that the latter
+example shows as partial representations.
+
+1. Let _item_ be the blob of data to hash.
+2. Let _hashList_ be an empty list.
+2. Foreach _(attr, value)_ pair in _item_:
+   1. If _value_ is null, continue.
```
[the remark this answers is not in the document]
arnau: That's a good question. Probably this belongs to a normalisation/canonicalisation/cleanup process. I'll make it explicit anyway.

content/item-hash/index.md
```diff
@@ -0,0 +1,257 @@
...
+Note that a Set is unordered at origin so order is applied as part of the
+hashing process.
+
+
+For reference, this is the abstract definition of an item:
+
+```elm
+type Value
+  = VString String    -- cardinality 1
+  | VSet (Set String) -- cardinality n
```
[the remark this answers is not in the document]
arnau: They are variants of the type `Value`. It is similar to the function of enums in some languages; a way to express multiple types as one (some times called union type).

content/item-hash/index.md
```diff
@@ -0,0 +1,257 @@
...
+```elm
+hashTagged : String -> Hash
+
+hashString : String -> Hash
+
+hashSet : Set -> Hash
+
+hashValue : Value -> Hash
+
+hashPair : (String, Value) -> List Hash
```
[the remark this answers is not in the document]
arnau: It is indeed a mistake, the correct one is `hashPair : (String, Value) -> Hash`. 👍

content/item-hash/index.md
```diff
@@ -0,0 +1,257 @@
...
+         2. Otherwise, normalise _el_ according to [string normalisation](#string-normalisation-algorithm)
+            tag it with `u` (String), hash it and append it to _elList_.
+      3. Concatenate _elList_ elements, tag it with `s` (Set), hash it and set
+         it to _valueHash_.
+   3. If _value_ starts with `**REDACTED**`, set _normValue_ with _value_
+      without `**REDACTED**`.
+   4. Otherwise, normalise _value_ according to [string normalisation](#string-normalisation-algorithm)
+      tag it with `u` (String), hash it and set _valueHash_.
+   5. Tag _attr_ with `u` (String), hash it and set _attrHash_.
+   6. Concat _attrHash_ and _valueHash_ and append to _hashList_.
```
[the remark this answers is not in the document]
michaelabenyohai: It looks like you made a change for this and later reverted it in another commit?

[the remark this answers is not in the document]
arnau: Did I, damn 😢

content/item-hash/index.md
michaelabenyohai: It doesn't look like this has changed. Does that mean it's correct, if so I'm not sure I understand.

content/item-hash/index.md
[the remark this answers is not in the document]
arnau: Re-reverted :)

content/item-hash/index.md
arnau: You are right, fixing it now

PR #26, opened 2018-08-13
RFC0013: Multihash

arnau:
### Context

Currently we use a custom `sha-256:` prefix for stringified hexadecimal hashes to qualify them as SHA256. With this approach parsers must be done specifically for registers and registers must standarise as many hashing algorithms as needed by users.

This RFC proposes to change the method to qualify hashes from using prefixes
`sha-256:` to using [Multihash](https://multiformats.io/multihash/) which is becoming a standard.

### Guidance to review

Check it is consistent.

### DIFF
--- a/content/multihash/index.md
+++ b/content/multihash/index.md
@@ -1,0 +1,64 @@
+---
+rfc: 0013
+start_date: 2018-08-13
+pr: openregister/registers-rfcs/pull/26
+status: draft
+---
+
+# Multihash
+
+## Summary
+
+This RFC proposes to change the method to qualify hashes from using prefixes
+`sha-256:` to using [Multihash](https://multiformats.io/multihash/).
+
+**This RFC is not backwards compatible in the same sense as
+[RFC0010](https://github.com/openregister/registers-rfcs/pull/24)**
+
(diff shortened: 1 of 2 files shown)


content/multihash/index.md
```diff
@@ -0,0 +1,64 @@
...
+stringified hexadecimal hashes to qualify them as SHA256. With this approach
+parsers must be done specifically for registers and registers must standarise
+any hashing algorithm needed by users. Relying on a standard helps benefiting
+from that standards ecosystem.
+
+## Explanation
+
+The proposal is to adopt [Multihash](https://multiformats.io/multihash/).
+Their specification is quite thorough but the gist is to replace any of our
+`sha-256:` to `1220`. `0x12` for the hashing function sha2-256 and `0x20` for
```
MatMoore: This confused me initially as I didn't realise you actually return the hex representation of the bytes as a string. I think you could leave out the hex representation here and just say "12" and "20".

arnau: Umm, I see your point, I would like a way to express it such that it's clear `12` is in hex and not base 10 or something else. Any ideas?

PR #27, opened 2018-08-17
RFC0014: Summary resource

arnau:
### Context

The current summary is exposed as the register resource (`GET /register`).
The name of the resource is misleading as you don't get the register, only
some metadata about it. Also, the current specification and implementation
don't agree on what is the register resource.

This RFC proposes a summary resource to offer an overview of the current state
of the register.

### Guidance to review

It should make sense as an independent resource and potentially an embeddable resource in paginated resources (probably non-normative).

### DIFF
--- a/content/summary-resource/index.md
+++ b/content/summary-resource/index.md
@@ -1,0 +1,101 @@
+---
+rfc: 0014
+start_date: 2018-08-13
+pr:
+status: draft
+---
+
+# Summary resource
+
+## Summary
+
+This RFC proposes a summary resource to offer an overview of the current state
+of the register.
+
+Dependencies:
+
+* [RFC0013: Multihash](https://github.com/openregister/registers-rfcs/pull/26)
+
+
+## Motivation
+
+The current summary is exposed as the register resource (`GET /register`).
+The name of the resource is misleading as you don't get the register, only
+some metadata about it. Also, the current specification and implementation
+don't agree on what is the register resource.
+
+## Explanation
+
+The summary resource is a new resource that will replace the old register
+resource. The summary resource will not contain any information about fields
+as this will be addressed by the schema resource.
+
+The most noticeable change is the addition of the `hashing-algorithm` object
+describing the hashing algorithm used in the register to create hashes for
+items, entries and proofs.
+
+***
+### Endpoint
+
+```
+GET /summary
+```
+
+### Response attributes
+
+|Name|Type|Description|
+|-|-|-|
+|`copyright`| String |The copyright of the data.|
+|`custodian`| Optional String |The custodian of the register.|
+|`description`| Optional String |The description of the register.|
+|`hashing-algorithm`| [HashingAlgorithm](#hahing-algorithm-attributes) |The hahing algorithm used throught the register.|
+|`last-updated`| Timestamp |The date the register was last updated.|
+|`licence`| String |The licence of the data.|
+|`total-entries`| Integer |The number of entries in the register.|
+|`total-items`| Integer |The number of items in the register.|
+|`total-records`| Integer |The number of records in the register.|
+
+#### Hashing algorithm attributes
+
+|Name|Type|Description|
+|-|-|-|
+|`digest-length`| Hexadecimal Integer |The length of the digest in bytes.|
+|`function-type`| Hexadecimal Integer |The identifier of the hash function.|
+|`name`| String |The common name of the algorithm.|
+
+See also: [Multihash table function type
+identifiers](https://github.com/multiformats/multihash/blob/master/hashtable.csv).
+
+***
+
+***
+**EXAMPLE:**
+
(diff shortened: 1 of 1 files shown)


content/summary-resource/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Response attributes
+
+|Name|Type|Description|
+|-|-|-|
+|`copyright`| String |The copyright of the data.|
+|`custodian`| Optional String |The custodian of the register.|
+|`description`| Optional String |The description of the register.|
+|`hashing-algorithm`| [HashingAlgorithm](#hahing-algorithm-attributes) |The hahing algorithm used throught the register.|
+|`last-updated`| Timestamp |The date the register was last updated.|
```
gidsg: Is it worth defining the time format of `last-updated` e.g. is it a Timestamp according to: https://openregister.github.io/specification/#timestamp-datatype ?

arnau: Sorry, yes all types are defined according to the spec.

arnau: This means that "String" is a string as defined by the specification too.

content/summary-resource/index.md
```diff
@@ -0,0 +1,101 @@
...
+items, entries and proofs.
+
+***
+### Endpoint
+
+```
+GET /summary
+```
+
+### Response attributes
```
gidsg: would we want to include the root hash in this response?

arnau: Yes, good call!

content/summary-resource/index.md
```diff
@@ -0,0 +1,105 @@
...
+
+|Name|Type|Description|
+|-|-|-|
+|`copyright`| String |The copyright of the data.|
+|`custodian`| Optional String |The custodian of the register.|
+|`description`| Optional String |The description of the register.|
+|`hashing-algorithm`| [HashingAlgorithm](#hahing-algorithm-attributes) |The hahing algorithm used throught the register.|
+|`root-hash`| Hash | The Merkle tree root hash. |
+|`last-updated`| Timestamp |The date the register was last updated.|
+|`licence`| String |The licence of the data.|
```
MatMoore: What is a valid string for this? Should we require an open source licence?

arnau: I think we can't afford to be restrictive here. OGL is one of many and these things change too often to be prescriptive at the spec level.

arnau:
> I think there are two parts to this RFC - adding the new object and renaming the endpoint.

Agree

> Adding the hashing-algorithm object makes sense, but I'm not enthusiastic about renaming the endpoint right now. Summary is slightly better but it still doesn't communicate to me what the endpoint is for, and it doesn't outweigh the inconvenience of a breaking change. Ideally I would like us to version the APIs and have a clear policy about how we deprecate APIs, but maybe this is a topic for another RFC.

The version strategy deserves an RFC on its own and it is likely to be a piece of work aside from the current Q2 one.

Regarding the rename, it's a soft change given that we can redirect 301 users but I'm happy to hold on this renaming and discuss it another time.

> The best name for this thing depends on what the intended use cases are for this endpoint and whether we have any plans to extend it.

Not plans per se, it might need evolution but none I can foresee right now.

> For example, as an API user I imagine I will want some way to retrieve metadata about the current schema, once that is available (not necessarily the full log of schema changes). At the moment it seems like you have to know about the register/datatype/field registers to do this (this is how the MOJ client does it) but this could potentially be returned from this endppoint.

See #28 The schema is distinct from the summary on purpose.

content/summary-resource/index.md
MatMoore: Just wondering if there's a way we can encourage some canonical naming so we don't end up with a bunch of registers using different variations of the same licence name.

arnau: We could add a non-normative section in the spec suggesting something on the lines of https://help.github.com/articles/licensing-a-repository/#searching-github-by-license-type

Or, we could consider having a register for known licenses. I'm not a 100% on board with this one because the custodianship model applied to this but it would be a very useful register to have. What do you think?

MatMoore: Something like that list would help I think.

I don't know if it should be register, but maybe the open standards team could comment on that. I think it would be useful if someone ever builds a catalog of reusable software/data across government, like [code.gov](https://code.gov)

arnau: This seems like a good list to base our restrictions https://spdx.org/licenses/

arnau: Mentioned in here https://github.com/alphagov/open-standards/issues/31

MatMoore: Cool - I think a non-non normative recommendation to use their identifiers would address my concern above.

arnau:
I'm rejecting this RFC in favour of RFC0018 (#31)

PR #28, opened 2018-08-17
RFC0015: Schema resource

arnau:
### Context

Before this RFC we explored a few options including JSON-LD, keep
something similar as the register register and field register or JSON
Schema. None of them is fit for purpose, mainly because our custom
datatype system.

This RFC proposes a schema resource to describe the register set of attributes
with their datatypes.

### Guidance to review

It is consistent and forwards compatible.


content/schema-resource/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+```
+GET /schema
+```
+
+### Parameters
+
+|Name|Type|Description|
+|-|-|-|
+|`log-size` | Optional `Integer`| Set the log size to the given entry number instead of the latest value.|
```
MatMoore: Not sure I understand this param, is it to retrieve the schema at a particular point in the log?

If so, it would be great if the records endpoint could behave in the same way eventually.

arnau: Yes, that's the purpose and I agree the data counterpart should have the same

arnau:
After @MatMoore comments here and in #27  and stepping back and thinking in context of #20 I'm making this RFC rejected because it's not fit for purpose.

PR #30, opened 2018-08-23
RFC0017: Full item redaction

arnau:
### Context

The specification defines “redaction” as:

> An item may be removed from a register, but the entry MUST NOT be removed
> from the register.

In some buried decision log and as a known thing for some components of the
Registers core team, a while ago it was decided that to fully redact an item,
the HTTP response would be a `410 Gone`. The specification doesn't reflect
this decision nor the details.

This RFC proposes a way to signal an item has been fully redacted.

Dependencies:

* [RFC0010: Item hash](https://github.com/openregister/registers-rfcs/pull/24).
* [RFC0013: Multihash](https://github.com/openregister/registers-rfcs/pull/26).

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ REST API).
 |0014|[Summary resource](content/summary-resource/index.md)|rejected|2018-08-24|#27|
 |0015|[Schema resource](content/schema-resource/index.md)|rejected|2018-08-22|#28|
 |0016|Blob resource|draft|-|#29|
-|0017|Full item redaction|draft|-|#30|
+|0017|[Full item redaction](content/full-item-redacted/index.md)|approved|2018-09-25|#30|
 |0018|Context resource|draft|-|#31|
 |0019|[Boolean datatype](content/boolean-datatype/index.md)|approved|2018-08-29|#32|
 |0020|[Blob normalisation](content/blob-normalisation/index.md)|approved|2018-09-07|#33|
--- a/content/full-item-redacted/index.md
+++ b/content/full-item-redacted/index.md
@@ -1,8 +1,9 @@
 ---
 rfc: 0017
 start_date: 2018-08-23
+decision_date: 2018-09-25
 pr: openregister/registers-rfcs#30
-status: draft
+status: approved
 ---
 
 # Full item redaction
--- a/content/full-item-redacted/index.md
+++ b/content/full-item-redacted/index.md
@@ -1,0 +1,90 @@
+---
+rfc: 0017
+start_date: 2018-08-23
+pr: openregister/registers-rfcs#30
+status: draft
+---
+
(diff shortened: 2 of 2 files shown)


MatMoore:
This broadly makes sense, but I think some differences between the redacted/non-redacted versions makes it harder for clients to talk to registers apis.

In your JSON example, the content type is application/json, but the content isn't:

```
HTTP/1.1 410 Gone
Content-Type: application/json

**REDACTED**12206b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb
```

So that would probably break client code. I think it would be valid json if it was just a string value like

```
HTTP/1.1 410 Gone
Content-Type: application/json

"**REDACTED**12206b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb"
```

We could also make fully redacted items a different content type to regular items, to indicate to clients how to handle the response. Would it make sense to create a vendored mime type like `application/vnd.bla+json`?

In your CSV example, I'm having trouble parsing this sentence: `For CSV, the payload MUST be the same as the expected list of attributes doesn't apply`. The payload must be the same as what?

This example doesn't have a header row, but a non-redacted item does have a header row. I think the fully redacted version should still have a header row, it just can't be the same as the normal one as there is only one value.

Also, this RFC focuses on items, but I don't think it explicitly says whether the `/records` endpoint should also return this 410 response.

arnau:
> In your JSON example, the content type is application/json, but the content isn't:

Thanks, fixed.

> We could also make fully redacted items a different content type to regular items, to indicate to clients how to handle the response. Would it make sense to create a vendored mime type like application/vnd.bla+json?

I'm not sure to what extent this helps users handle this payload better. Let me think about it.

> In your CSV example, I'm having trouble parsing this sentence

It makes no sense. I was thinking in the best way to approach this problem and got two solutions mixed :(

As you highlight later on, the non-redacted version has headers. Given that the data is completely different and it is no longer a 200, I'm seriously considering to stick to JSON only even when the user asks for CSV. In terms of content negotiation this is expected so it shouldn't be too bad.

I'm open to alternatives, one option could be to have a structure for errors that can be serialised as JSON or CSV.

> Also, this RFC focuses on items, but I don't think it explicitly says whether the /records endpoint should also return this 410 response.

Very good point. The record should be 200 and show the redacted item inline as with the rest. I'll make it explicit 👍

arnau:
I have thought about this and now I'm more convinced 410 doesn't need a payload because it doesn't really provide anything on top of what HTTP provides. I'll update this RFC accordingly.

arnau:
Further thinking made me realise the 410 solution is broken and it's better to solve this problem building on top of RFC0010. The RFC explains that now.

Issue #17, opened 2018-05-22
How could registers be mothballed?


nacnudus:
### Context

* RFC code:

### Explanation

So far all the registers are relevant and have active custodians.  There are two (or more?) ways this could change:

* A custodian could stop maintaining the register
* A register could become obsolete because the state of the world changes, e.g. if allergens are no longer regulated
* The scope of a register could change, e.g. if powers are devolved

Given that registers won't ever be deleted, they ought to declare whether they are currently valid or not.  How to do this?

#### End-date all the records

A possible workaround, but it implies that all the things in the register no longer exist.

#### Tombstone all the records

We don't use tombstoning for anything yet.

#### Give every record equal start-date and end-dates, with 1-second precision

This is a current workaround for invalidating records, and we don't like it.

#### Invalidate all records

When we have implemented this.  It's a bit too strong -- the records *were* valid.

#### End-date the record in the register register

The register register doesn't have start-date and end-date fields.  Even if it did, users of registers would have to look here for the state of any register.

#### Set the custodian and registry to empty strings

This is a bit weak.  It might be something to do *as well as* mothball the register.

arnau:
My gut feeling is that the register should be able to express it's no longer active (ended, closed, …) at the register metadata level, not at the record level.

PR #31, opened 2018-08-23
RFC0018: Context resource

arnau:
### Context

Currently, the register metadata is not exposed consistently. Some is exposed
via the register resource, some via the field register and some from the
register register.

This RFC proposes a context resource to surface the metadata of the register.

Meanwhile there is no versioning defined for metadata treat this as single
versioned.

Dependencies:

* [RFC0013: Multihash](https://github.com/openregister/registers-rfcs/pull/26)

Note: It obsoletes #27 

Addresses #17

### DIFF
--- a/content/context-resource/index.md
+++ b/content/context-resource/index.md
@@ -1,0 +1,325 @@
+---
+rfc: 0018
+start_date: 2018-08-13
+pr: openregister/registers-rfcs#31
+status: draft
+---
+
+# Context resource
+
+## Summary
+
+This RFC proposes a context resource to surface the current metadata of the
+register.
+
+Dependencies:
+
+* [RFC0013: Multihash](../multihash/index.md)
+
+
+## Motivation
+
+Currently, the register metadata is not exposed consistently. Some is exposed
+via the register resource, some via the field register and some from the
+register register.
+
+Also, the current specification and implementation don't agree on what is the
+register resource.
+
+The main driver for this RFC is to have a computed resource for the latest
+metadata in a way that makes the register self contained (i.e. a user should
+not need to interact with any other register to get the metadata for a
+register).
+
+
+## Explanation
+
+The **context** is the metadata snapshot that apply to a given log size.
+
+### Definition
+
+```elm
+type Context =
+  { id : Name
+  , copyright : Maybe String
+  , custodian : Maybe String
+  , description : Maybe String
+  , hashingAlgorithm : HashingAlgorithm
+  , licence : Maybe String
+  , rootHash : Hash
+  , schema : Schema
+  , stats : Stats
+  , status : Status
+  , title : Maybe String
+  }
+```
+
+See each section below of the explanation for each attribute.
+
+
+#### Id
+
+* Type: Name (This is defined in the specification as `[a-z][a-z0-9-]*`, the
+  future specification will make it more clear).
+
+The register identifier. The current implications for this value make it act
+as a unique identifier for the environment it belongs to. This RFC doesn't
+pretend to change this fact.
+
+
+#### Copyright
+
+* Type: Optional String
+
+The register copyright. E.g. `© Crown copyright`. Similar to the `copyright`
+field in the Register register. Must be present for any published register.
+
+
+#### Custodian
+
+* Type: Optional String
+
+The same value currently exposed in the Register resource. Must be present for
+any published register.
+
+
+#### Description
+
+* Type: Optional String
+
+The human readable description of the register.
+
+
+#### Hashing algorithm
+
+* Type: HashingAlgorithm
+
+The hashing algorithm used throughout the register.
+
+```elm
+type HashingAlgorithm =
+  { id : Name
+  , functionType : UVarint
+  , digestLength : Int
+  }
+```
+
+This depends on RFC0013
+
+|Name|Description|
+|-|-|
+|`digest-length`|The byte length of the digest. E.g. `0x12`|
+|`function-type`|The identifier of the hash function. E.g. `0x20`, `0xb220`|
+|`id`| The name of the algorithm. E.g. `sha2-256`|
+
+See also: [Multihash table function type
+identifiers](https://github.com/multiformats/multihash/blob/master/hashtable.csv).
+
+
+#### Licence
+
+* Type: Optional String
+
+The licence that applies to the data. Must be present for any published register.
+
+This could get restricted with a suitable list of licences and validated with
+an external tool such that the technology allows free text but each register
+controller is able to ensure what is allowed.
+
+
+#### Root hash
+
+* Type: Hash
+
+The root hash for the register. Note that signatures will be addressed in
+another RFC. This is not strictly part of the metadata, it is derived from the
+data.
+
+#### Schema
+
+* Type: Schema
+
+The set of attributes that define the data allowed in the register.
+
+```elm
+type Attribute =
+  { id : Name
+  , datatype : Datatype
+  , title : Maybe String
+  , description : Maybe Text
+  }
+
+type Schema =
+  Set Attribute
+```
+
+|Name|Description|
+|-|-|
+|Id|Attribute identifier (e.g. `start-date`)|
+|Datatype|Datatype identifier (e.g. `datetime`)|
+|Title|Human readable attribute name (e.g. `Start date`)|
+|Description|Human readable attribute description (e.g. `The date a record first became relevant to a register.`)|
+
+
+#### Stats
+
+* Type: Stats
+
+The summary of objects stored in the register. This overlaps with the Register
+resource and it is not strictly part of the metadata, it is derived from the data.
+
+
+#### Status
+
+* Type: Status (Active, Retired). Defaults to Active.
+
+The status of the register. Either active or retired. This addresses the
+problem of being able to retire an entire register (see
+[issue #17](https://github.com/openregister/registers-rfcs/issues/17)).
+
+```elm
+type Status
+  = Active { startDate : Timestamp }
+  | Retired { startDate : Timestamp
+            , endDate : Timestamp
+            , replacement : Maybe Url
+            , reason : Text
+            }
+```
+
+|Name|Description|
+|-|-|
+|Start date| Date when the register started.|
+|End date| Date when the register was retired.|
+|Replacement| A URL to a register that replaces the current one.|
+|Reason| A human readable explanation of why the register is no longer active.|
+
+
+#### Title
+
+* Type: Optional String
+
+The human readable name of the register.
+
+
+### HTTP resource
+
+***
+#### Endpoint
+
+```
+GET /context
+```
+
+#### Parameters
+
+|Name|Type|Description|
+|-|-|-|
+|`total-entries` | Optional `Integer`| The log size to compute the snapshot from. Note this only applies when the metadata can be versioned.|
+
+
+#### Response attributes
+
+|Name|Type|
+|-|-|
+|`id`| Name |
+|`copyright`| Optional String |
+|`custodian`| Optional String |
+|`description`| Optional String |
+|`hashing-algorithm`| HashingAlgorithm |
+|`licence`| Optional String |
+|`root-hash`| Hash |
+|`schema`| List Attribute |
+|`stats`| Stats |
+|`status`| Status |
+|`title`| Optional String |
+
+#### Hashing algorithm attributes
+
+|Name|Type|
+|-|-|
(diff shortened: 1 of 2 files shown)


content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+
+See each section below of the explanation for each attribute.
+
+
+#### Id
+
+* Type: Name (This is defined in the specification as `[a-z][a-z0-9-]*`, the
+  future specification will make it more clear).
+
+The register identifier. The current implications for this value make it act
```
MatMoore: What do you mean by "current implications"?

I'm not sure if this makes sense, but I think the environment it belongs to is still defined by the register or register (or another catalog of registers), which makes me wonder whether all of this metadata actually needs to be exposed through the API.

Things like title, description and licence seem very catalogue-y in the sense that they are relevant when deciding whether to use a register, but if you're already consuming from one then it doesn't seem important.

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+as a unique identifier for the environment is belongs to. This RFC doesn't
```
MatMoore: small typo - s/is/it

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+pretend to change this fact.
+
+
+#### Copyright
+
+* Type: String
+
+The register copyright. E.g. `© Crown copyright`. Similar to the `copyright`
+field in the Register register but required to be informed.
```
MatMoore: what does required to be informed mean?

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+
+
+#### Custodian
+
+* Type: String
```
MatMoore: You have this as a maybe above, is this intended to represent the case mentioned in issue #17 where a register loses its custodian?

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+
+The same value currently exposed in the Register resource.
+
+
+#### Description
+
+* Type: Optional String
```
MatMoore: Any reason why this can't be mandatory if we return it here? A register should always have a human readable description, otherwise how do people know to use it?

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+type Schema =
+  Set Attribute
+```
+
+|Name|Description|
+|-|-|
+|Id|Attribute identifier (e.g. `start-date`)|
+|Datatype|Datatype identifier (e.g. `datetime`)|
```
MatMoore: Does this still have to refer to a record in the datatype register?

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+data.
+
+#### Schema
+
+* Type: Schema
+
+The set of attributes that define the data allowed in the register.
+
+```elm
+type Attribute =
```
MatMoore: Does this eliminate the need for the field register?

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+The status of the register. Either active or retired. This addresses the
+problem of being able to retire an entire register (see
+[issue #17](https://github.com/openregister/registers-rfcs/issues/17)).
+
+```elm
+type Status
+  = Active { startDate : Timestamp }
+  | Retired { startDate : Timestamp
+            , endDate : Timestamp
+            , replacement : Maybe Url
```
MatMoore: I think this is the first time registers have linked to other registers by URL.

This could cause problems if that URL breaks but the register is still available elsewhere.

Another way of doing this would be to use a CURIE then rely on a catalog to resolve this to something (and this could be a catalog outside of GDS's control if GDS is no longer hosting the register register)

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+  ]
+}
+```
+
+***
+
+### Considerations
+
+The context resource makes the register resource obsolete. Given that users
+depend on it the register resource will be available for as long as necessary.
```
MatMoore: How long is that? This shouldn't need to apply to new registers, and it would be great if we didn't have to support it in the reference implementation forever as well.

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+    {"id": "name", "datatype": "name", "cardinality": "1"},
+    {"id": "function-type", "datatype": "integer", "cardinality": "1"},
+    {"id": "digest-length", "datatype": "integer", "cardinality": "1"}
+  ]
+}
+```
+
+***
+
+### Considerations
```
MatMoore: Does this have any implications for what will happen to the register register? Is there a mechanism to ensure the register register stays in sync with what the registers themselves report for this metadata?

content/context-resource/index.md
```diff
@@ -0,0 +1,320 @@
...
+status: draft
+---
+
+# Context resource
+
+## Summary
+
+This RFC proposes a context resource to surface the metadata of the register.
+
+Meanwhile there is no versioning defined for metadata treat this as single
```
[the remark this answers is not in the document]
arnau: Not my best moment, I'll admit it :D 

It accounts for the current situation where there is no schema evolution so the first version is the only version (single).

content/context-resource/index.md
MatMoore: Fair enough 👍

content/context-resource/index.md
arnau: The goal is to make a register self descriptive. Personally, I would like to not have a unique ID for a register within a register but it is how things are right now.

On title, description, again the goal is to have a self contained register such that it can be transported with minimal lose.

The catalogue (register register) right now duplicates this information and it is something we will need to address in the future.

content/context-resource/index.md
arnau: It means it is not optional. A register must have an explicit copyright

content/context-resource/index.md
arnau: Yes sorry, I'll amend this, it needs to be `Optional String`.

content/context-resource/index.md
arnau: Well, ideally yes and it is something we will probably want to enforce for registers part of the our catalogue but I can foresee registers not having a description at early/immature stages.

content/context-resource/index.md
arnau: The idea is to get to the point where the datatype register is either obsolete or complementary to the spec. So the answer to your question is no, it needs to refer to a valid datatype identifier as specified in the spec.

content/context-resource/index.md
arnau: I agree, I think this will be answered next quarter when we define the version strategy, but I'm open to suggestions

content/context-resource/index.md
arnau: Nothing concrete just yet, but we have to think about this soon.

content/context-resource/index.md
```diff
@@ -0,0 +1,325 @@
...
+type Context =
+  { id : Name
+  , copyright : Maybe String
+  , custodian : Maybe String
+  , description : Maybe String
+  , hashingAlgorithm : HashingAlgorithm
+  , licence : Maybe String
+  , rootHash : Hash
+  , schema : Schema
+  , stats : Stats
```
gidsg: can we make this `statistics`? I would rather we did not use abbreviations for clarity.

content/context-resource/index.md
```diff
@@ -0,0 +1,325 @@
...
+
+```elm
+type Attribute =
+  { id : Name
+  , datatype : Datatype
+  , title : Maybe String
+  , description : Maybe Text
+  }
+
+type Schema =
```
gidsg: This seems like what we refer to in other place as `fields` is there a deliberate attempt to move away from this to schema? If so we should be consistent across other instances e.g. RSF. if not, we should call it `fields`

content/context-resource/index.md
gidsg: Right now I cannot see where this information would be persisted? Is there a companion to this that addresses how this is represented in RSF/Entry log? If not I think this should be pulled out into a separate RFC that describes Status and its representation.

content/context-resource/index.md
```diff
@@ -0,0 +1,325 @@
...
+|`cardinality`| Cardinality |
+|`title`| Optional String |
+|`description`| Optional String |
+
+#### Stats attributes
+
+|Name|Type|
+|-|-|
+|`total-entries`| Integer |
+|`total-items`| Integer |
```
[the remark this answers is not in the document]
gidsg: I am not sure, can you think of a use case for `total-items` @arnau?

content/context-resource/index.md
arnau: @MatMoore I did the research on using json schema or csvw schemas and json-ld related and the result is that Registers is not really 100% compatible with them due our primitives. Both csvw and json-schema let you define your own patterns which could work for us as a sort-of compatible translation but it's messier than it seems.

Now, particularly with csvw, it allows to provide schema + context although probably not everything this context resource is offering. I'm not sure it's worth it extracting (again) the schema just because of that.

@gidsg It is deliberated indeed. The "fields" approach builds on a **central** vocabulary held in the fields register and it has been proven a mistake. Moving away from this central model to a local schema definition predates me, but I totally agree with it.

Changing terminology to schema and attributes is important because a) it's not the same and b) when it is (i.e. field vs attribute) I want to be able to talk about previous concepts with minimal confusion.

RSF is something we can change when we are ready and comfortable but to my eyes it is not important enough to reject or hold this RFC. After all, RSF is an internal format that so far has not even a stable specification.

content/context-resource/index.md
```diff
@@ -0,0 +1,325 @@
...
+* Type: Optional String
+
+The human readable description of the register.
+
+
+#### Hashing algorithm
+
+* Type: HashingAlgorithm
+
+The hashing algorithm used throughout the register.
```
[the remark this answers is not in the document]
arnau: I think it's a nice thing to offer and not at a big cost. Why do you think it shouldn't be there? The fact that you can infer it doesn't mean it is useless or harmful.

content/context-resource/index.md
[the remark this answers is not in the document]
arnau: I added it for consistency, in all fairness, I had my doubts about the usefulness of the other totals. But, if we agree these are useful, then items is useful as well, items (blobs) are as important and relevant as entries at the very least and treating them as second citizens don't help anyone. Is there any other argument for not having them? is it harmful in any way?

arnau:
> I'm also wondering whether this would prevent us from supporting another metadata standard.
>
> For example, if we wanted to provide things like csvw or json schema it would be great if we just had one resource and you can request the standard you want, using content types. But I'm not sure if that would fit this model where the schema is part of a larger object.

You mentioned three times csvw and json schema so I think it's worth talking about it.

When I started looking for ways to express the schema I started with known standards like json schema and csvw metadata. As it turns out Registers define their own primitive datatypes that are not compatible with XML Schema which is what csvw is based on and it is not compatible either with the defaults of json schema. So, in order to adhere to any of these we have to either express the datatypes in a pattern-oriented manner or change the datatypes we have (I'm ruling this out unless there is no other choice).

I might have missed something essential from json schema or csvw or json-ld contexts that could be used instead of the proposal. Happy to revisit if you have found it.

content/context-resource/index.md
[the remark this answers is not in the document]
MatMoore: It's not harmful, I guess my thinking is that an API users doesn't really care how many items there should be, as long as they can link every entry to an item. We also don't have a list endpoint for items like we do for records and entries so I don't see think its inconsistent to not count them.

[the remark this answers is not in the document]
arnau: But we will have the an `/items` (`/blobs`) endpoint, the ADR is there and the spec defines it.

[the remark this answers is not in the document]
MatMoore: Ah, my mistake. In that case ignore this comment

content/context-resource/index.md
gidsg: @arnau please can you address this [comment](https://github.com/openregister/registers-rfcs/pull/31/files#r225571551)?

arnau: @gidsg status must be part of the register metadata which means it has to be encoded in RSF. The way to express that will highly depend on the metadata solution we end up having but, if it is in line with what we have right now it will be an system entry with the appropriate key pointing to a blob with these information.

PR #32, opened 2018-08-23
RFC0019: Boolean datatype

arnau:
### Context

The datatypes in registers cover most of the common values you find in other
systems. Although some systems opt to represent booleans as 0 and 1, it is
ambiguous without human interaction.

This RFC proposes a new datatype to represent boolean values.

### DIFF
--- a/content/boolean-datatype/index.md
+++ b/content/boolean-datatype/index.md
@@ -1,9 +1,9 @@
 ---
 rfc: 0019
 start_date: 2018-08-23
-decision_date:
+decision_date: 2018-08-29
 pr: openregister/registers-rfcs#32
-status: draft
+status: approved
 ---
 
 # Boolean datatype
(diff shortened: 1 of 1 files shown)


content/boolean-datatype/index.md
```diff
@@ -0,0 +1,52 @@
...
+boolean = "true" / "false"
+```
+
+***
+**EXAMPLE:**
+
+For example, in JSON:
+
+```json
+{"foo": "true", "bar": "false"}
```
gidsg: json does have a native boolean type, but I guess we are deliberately using a string here?

arnau: We are indeed. Items are not fully fleshed JSON in any other datatype. All datatypes are consistently represented as strings. This ensures parity between leaf values and between serialisation formats.

PR #33, opened 2018-08-30
RFC0020: Blob normalisation

arnau:
### Context

The current specification _suggests_ how null values and empty strings should
be treated in terms of normalisation but it is not clear so it is open for
interpretation.

This RFC proposes an algorithm to normalise blobs of data such that the
hashing algorithm (#24) is predictable.

### DIFF
--- a/content/blob-normalisation/index.md
+++ b/content/blob-normalisation/index.md
@@ -1,0 +1,71 @@
+---
+rfc: 0020
+start_date: 2018-08-30
+pr: openregisters/registers-rfc#33
+status: draft
+---
+
+# Blob normalisation
+
+## Summary
+
+This RFC proposes an algorithm to normalise blobs of data such that the
+hashing algorithm (RFC0010) is predictable.
+
+Note that throughout this RFC _blob_ and _item_ are used interchangeably.
+
+
+## Motivation
+
+The current specification _suggests_ how null values and empty strings should
+be treated in terms of normalisation but it is not clear so it is open for
+interpretation.
+
+## Explanation
+
+Blobs of data must have a normal form such that algorithms processing them
+(e.g. hashing algorithm (RFC0010)) have a consistent output. Blobs of data are
+attribute-value pair associative arrays where values are Strings but nulls
+have to be addressed with care.
+
+The following algorithm ensures all nullable values are removed from the
+normalised representation:
+
+1. Let _blob_ be the blob to normalise.
+2. Let _result_ be an empty dictionary.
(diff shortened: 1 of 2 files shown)


content/blob-normalisation/index.md
```diff
@@ -0,0 +1,71 @@
...
+The following algorithm ensures all nullable values are removed from the
+normalised representation:
+
+1. Let _blob_ be the blob to normalise.
+2. Let _result_ be an empty dictionary.
+3. Foreach _(attr, value)_ pair in _blob_:
+   1. If _value_ is null, continue.
+   2. If _value_ is an empty String, continue.
+   3. If _value_ is an empty Set, continue.
+   4. If _value_ is a Set:
```
MatMoore: When we say Set here, do we mean just "a cardinality N field"?

I think using objecthash types to describe this algorithm could be confusing. It makes it sound more generic than it is, since there are other types like List and Map could be read as falling into the "otherwise...", but there's an implicit assumption that they won't be part of a blob.

arnau:
> Should this normalisation include sorting attribute names lexicographically?

No need for that since the new item hashing algorithm doesn't require preordering data.

> Other than that, this makes sense to me, although can we say up front that this will replace the previous canonicalisation algorithm?

I think we can, yes 👍

content/blob-normalisation/index.md
arnau: Well, the normalisation is for blobs and these exclusively allow strings and sets of strings. It's not that implicit if you know the definition of blob which hasn't changed since before my time in the team.

arnau:
In all fairness, the RFC that makes the canonicalisation obsolete is RFC0010 but I'll mention it in this one as well.

content/blob-normalisation/index.md
MatMoore: That makes sense, the problem for me is just that there are multiple vocabularies in use, so I'm not clear about what should be assumed and what should be defined at the top of the RFC.

I think there's room for confusion because items in the current spec are presented in JSON form, with no reference to sets at all, other than the unrelated "infoset" term (the examples show JSON arrays and describe them as cardinality-N fields).

The "strings and sets of strings" definition comes from the rewritten version (https://clever-sammet-9dbaef.netlify.com), which I don't think we're linking to publicly yet, but as this will be the basis for the next version, should I be assuming everything within in it is common knowledge for the purposes of reviewing RFCs?

Even then, our data model still refers to both `cardinality N strings` and `sets of strings`, and set isn't listed as a datatype. I think we should have a preferred term and stick to it 😄

PR #35, opened 2018-09-27
RFC0021: Archive resource

arnau:
**UPDATE**: After @MatMoore question on the actual purpose of the archive resource, we agreed in removing it. The new RFC reflects this decision.

### DIFF
--- a/content/archive-resource/index.md
+++ b/content/archive-resource/index.md
@@ -82,7 +82,7 @@ GET /archive?from={n}&to={n}
 Parameters:
 
 |Name|Type|Description|
-|-|-|
+|-|-|-|
 |from|Integer|The log size to use as starting point for the slice|
 |to|Integer|The log size to use as ending point for the slice|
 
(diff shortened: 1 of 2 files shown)


MatMoore:
I don't think I actually understand the use case for downloading entries and items as files in a zip. And also, from what I can see in google analytics this endpoint gets no traffic at the moment.

What if we just deprecated the endpoint instead?

arnau:
@MatMoore If I understand correctly where this endpoint comes from, it is the original way to get a full copy for a register in a single request. Internally, instead, we use RSF but RSF is not part of the specification and so far the position is that it shouldn't be standardised.

So, if we remove this endpoint we accept that there is no relevant user need to get a copy for the register in a single request, which according to the metrics you mention is the case. I'm happy to discuss in person, perhaps with @gidsg so we have all points of view on the matter.

MatMoore:
Sounds good 👍

content/archive-resource/index.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Summary
+
+This RFC proposes removing the Archive resource to simplify the API.
+
+This is a breaking change.
+
+## Motivation
+
+Currently, the specification defines the Archive resource as a zip file withi
```
MatMoore: there's a typo here

PR #39, opened 2018-09-28
RFC0022: Facets endpoint

arnau:
### Context

Currently the REST API for facets uses a path structure that overlaps with the
standard record structure. This makes path handling (e.g. analytics, logging)
harder than it needs to be. Observation suggests that users can get confused
of what is the resource about.

This RFC proposes to change the endpoint for the records' “facets” to simplify
path handling.

### DIFF
--- a/content/facets-endpoint/index.md
+++ b/content/facets-endpoint/index.md
@@ -1,9 +1,9 @@
 ---
 rfc: 0022
 start_date: 2018-09-28
-decision_date:
+decision_date: 2018-11-30
 pr: openregisters/registers-rfc#39
-status: draft
+status: approved
 ---
 
 # Facets endpoint
--- a/content/facets-endpoint/index.md
+++ b/content/facets-endpoint/index.md
@@ -50,7 +50,7 @@ records collection.
 GET /records{?name, value}
 ```
 
-Where `{?name}` is required if `{?value}` is informed.
+Where `{?name}` is required if `{?value}` is provided.
 
 
 ### Filter by strict value match
@@ -67,7 +67,7 @@ Exactly the current behaviour.
 GET /records?name={attribute-name}
 ```
 
-If the value of the given attribute name is informed (i.e. is not null) it
+If the value of the given attribute name is provided (i.e. is not null) it
 returns the record.
 
 ***
--- a/content/facets-endpoint/index.md
+++ b/content/facets-endpoint/index.md
@@ -1,0 +1,93 @@
+---
+rfc: 0022
+start_date: 2018-09-28
+decision_date:
+pr: openregisters/registers-rfc#39
+status: draft
+---
+
+# Facets endpoint
+
+## Summary
+
+This RFC proposes to change the endpoint for the records' “facets” to simplify
+path handling.
+
+## Motivation
+
+Currently the REST API for facets uses a path structure that overlaps with the
+standard record structure. This makes path handling (e.g. analytics, logging)
+harder than it needs to be. Observation suggests that users can get confused
+of what is the resource about.
+
+
+## Explanation
+
+For context, the current endpoint for facets is:
+
+```
+GET /records/{attribute-name}/{attribute-value}
+```
+
+The record endpoint is:
+
+```
+GET /records/{key}
+```
+
+And the records endpoint is:
+
+```
+GET /records
+```
+
+
+This RFC proposes to change the facets endpoint to extend the records endpoint
+with a query string to make more obvious that a “facet” is filter on the
+records collection.
+
+```
+GET /records{?name, value}
+```
+
+Where `{?name}` is required if `{?value}` is informed.
+
+
+### Filter by strict value match
+
+```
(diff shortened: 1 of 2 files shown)


content/facets-endpoint/index.md
```diff
@@ -0,0 +1,93 @@
...
+
+This RFC proposes to change the facets endpoint to extend the records endpoint
+with a query string to make more obvious that a “facet” is filter on the
+records collection.
+
+```
+GET /records{?name, value}
+```
+
+Where `{?name}` is required if `{?value}` is informed.
```
MatMoore: 'Informed' isn't the right word here - I think it should say something like "if `{?value}` is provided"

content/facets-endpoint/index.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+### Filter by strict value match
+
+```
+GET /records?name={attribute-name}&value={attribute-value}
```
MatMoore: A limitation of this is we can only filter using one attribute at a time, are we ok with that?

Also many registers have a "name" field so I don't know if it would be confusing to use `name` as a param here.

Another alternative would be something like

```
/records?filter_{attribute-name}={attribute-value}
```

Although you'd need to use a special value to implement the next use case.

IMO GOV.UK's search API implemented this quite well - their [filtering API](https://docs.publishing.service.gov.uk/apis/search/faceted-search.html#filtering-parameters) can either include or exclude stuff, and it can also express null values and date ranges.

MatMoore:
Have added a few comments about the details but I agree with the proposal itself.

content/facets-endpoint/index.md
arnau: > A limitation of this is we can only filter using one attribute at a time, are we ok with that? I'm unsure at this point what we want to do with facets, and whether we would use them to address any of the use cases indexes were designed to address.

The proposal changes the way we interact with facets but it doesn't change the capabilities. Accepting multiple attributes is a deep change that require much more than a change in the API.

> Also many registers have a "name" field so I don't know if it would be confusing to use name as a param here.

I don't agree with being confusing (or more or less confusing than any other combination like the one you suggest). But it's just my opinion. I don't  have proper backup to sustain it beyond that, do you?

> IMO GOV.UK's search API implemented this quite well - their filtering API can either include or exclude stuff, and it can also express null values and date ranges.

I'm going to disagree here. GOV.UK has a different need, and it looks like they went for something on the lines of the Elasticsearch DSL which is overkill for us.

content/facets-endpoint/index.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+Exactly the current behaviour.
+
+### Filter by name
+
+```
+GET /records?name={attribute-name}
```
gidsg: to clarify, I think this is new functionality there is no endpoint for this in the current API. I guess there is a tension with the filtering capabilities we should provide v.s. what clients should implement (as we saw with the indexes functionality). So I am not sure where we should be going with this but this particular one does feel logical as it feels like a subset of the facetted value query, just allowing for non-null. 
If there is more debate about this one option would be to move this new query into a separate RFC and make this one just about the new route/redirects.

content/facets-endpoint/index.md
```diff
@@ -0,0 +1,93 @@
...
+If the value of the given attribute name is provided (i.e. is not null) it
+returns the record.
+
+***
+**EXAMPLE:**
+
+For example, to get all records with an end date in the country register:
+
+```http
+GET /records?name=end-date HTTP/1.1
```
gidsg: I wonder if the naming is not great i.e. many registers have a field called `name` and users may confuse this functionality with something to do with the name value. Maybe `field-name` would be less ambiguous?

content/facets-endpoint/index.md
MatMoore: Just to clarify, I agree with the motivation for this RFC, my concern is just that the choice of URL structure will bake in certain constraints, and we won't be able to add certain kinds of functionality without making another breaking change later on.

Whether this is a big deal or not depends on how likely we are to actually do these kinds of changes, and how happy we would be to make breaking changes in this area of the API. I'm lacking the bigger picture here so I'm happy to defer to your judgement on this.

I don't think we need to split the RFC unless we want to have a bigger discussion about the facets functionality now?

gidsg:
I am not too worried about the constraints of this approach after all, nothing stops us adding say a `filter` parameter with other capabilities in future.

content/facets-endpoint/index.md
arnau: I have removed this part of the RFC to handle the null vs non-null conversation separated from the core intention of this proposal.

content/facets-endpoint/index.md
arnau: I don't think `name` is more or less ambiguous than `field-name`, `fieldname`, `attribute-name`, `column` or any other label. Just because `name` is used as a name doesn't seem strong enough to change to something else that may be used as well as an attribute name in a register.

Unless there are other arguments against it I'd rather keep it short.

PR #40, opened 2018-10-01
RFC0023: Proof resource

arnau:
### Context

Currently, the specification define the Proof resource endpoints parametrised
by type of proof through the `proof-identifier` parameter. The implications of
implementing a proof system are substantial as demonstrated by the long time
is taking the reference implementation to have a stable implementation for a
SHA2-256 Merkle tree. Also, a Register uses a single hashing function which
means the proof type has to align with that fact.

This RFC proposes to fix the type of proofs to Certificate Transparency's Merkle tree approach.

### DIFF
--- a/content/proof-resource/index.md
+++ b/content/proof-resource/index.md
@@ -1,9 +1,9 @@
 ---
 rfc: 0023
 start_date: 2018-10-01
-decision_date:
+decision_date: 2018-10-31
 pr: openregister/registers-rfcs/pull/40
-status: draft
+status: approved
 ---
 
 # Proof resource
--- a/content/proof-resource/index.md
+++ b/content/proof-resource/index.md
@@ -1,0 +1,174 @@
+---
+rfc: 0023
+start_date: 2018-10-01
+decision_date:
+pr: openregister/registers-rfcs/pull/40
+status: draft
+---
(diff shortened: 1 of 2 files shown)


content/proof-resource/index.md
```diff
@@ -0,0 +1,174 @@
...
+  "root-hash": "12208d92e1e0af1d43c41e498e6baed0d0b3ea2770d1bf9d2afc04e9c4dad7795729",
+  "total-entries": 208
+}
+```
+***
+
+#### Consistency proof
+
+```
+GET /proofs/consistency/{small-log-size}/{large-log-size}
```
MatMoore: Am I right in thinking that the most common use case will be for `large-log-size` to be the current size of the log? If so, can we make it optional?

arnau: I think it is indeed the most common use case. Do you think we need to address (agree?) on the optional bit in this RFC or are you comfortable to accept it as it is now, and address the implications for an optional bit in another RFC?

MatMoore: Either way is fine by me 👍

arnau: Ook, I'll merge then 👍

PR #41, opened 2018-10-02
RFC0024: Indexes removal

arnau:
### Context

After a period of assessment the Registers team concluded that indexes should
not be approached the way they are. As a consequence of this decision, we have
the opportunity to simplify the data model and the REST API, mostly roll back
to the state it was before the indexes work was introduced.

This RFC proposes the changes to make to the data model and REST API
consequence of removing the general concept of indexes.

This RFC is a breaking change.

This RFC supersedes [RFC0009](Entry hash) due the changes on the data model
that affect entries.

### DIFF
--- a/content/indexes-removal/index.md
+++ b/content/indexes-removal/index.md
@@ -1,9 +1,9 @@
 ---
 rfc: 0024
 start_date: 2018-10-02
-decision_date:
+decision_date: 2018-10-25
 pr: openregister/registers-rfcs#41
-status: draft
+status: approved
 ---
 
 # Indexes removal
--- a/content/indexes-removal/index.md
+++ b/content/indexes-removal/index.md
@@ -65,7 +65,7 @@ changing the entry hash (RFC0009) algorithm as follows:
 5. Let _blob_ be the list of bytes of the blob hash.
 6. Apply the _hashValue_ function to _number_ tagged with `0x69` (Integer). And
    append the result to _hashList_.
-7. Apply the _hashValue_ function to _key_ tagged with `0x74` (String). And
+7. Apply the _hashValue_ function to _key_ tagged with `0x75` (String). And
    append the result to _hashList_.
 8. Apply the _hashValue_ function to _timestamp_ tagged with `0x74` (Timestamp). And
    append the result to _hashList_.
--- a/content/indexes-removal/index.md
+++ b/content/indexes-removal/index.md
@@ -1,0 +1,182 @@
+---
+rfc: 0024
+start_date: 2018-10-02
+decision_date:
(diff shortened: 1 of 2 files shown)


content/indexes-removal/index.md
```diff
@@ -0,0 +1,182 @@
...
+### Data model
+
+The main change on the data model is the entry. It changes from:
+
+```elm
+type Entry =
+  { number: Integer
+  , timestamp: Timestamp
+  , key: ID
+  , item: Set Hash
```
MatMoore:
```suggestion
  , blob: Set Hash
```

content/indexes-removal/index.md
```diff
@@ -0,0 +1,182 @@
...
+|-|-|-|
+|`key`| [ID](/glossary/key#id-type)|The record identifier.|
+
+Response attributes:
+
+|Name|Type|Description|
+|-|-|-|
+|`entry-number`|Integer|The entry number.|
+|`entry-timestamp`| Timestamp|The entry timestamp.|
+|`key`| ID|The entry key.|
```
MatMoore: Is the key necessary in the response? It duplicates information from the URL

content/indexes-removal/index.md
```diff
@@ -0,0 +1,182 @@
...
+
+##### List records
+
+```
+GET /records
+```
+
+Summary:
+
+* Returns an array of record objects instead of an object with keys and
```
MatMoore: I think the object with keys actually makes sense here, because there is no need for the collection to be ordered, and it makes sense to be able to look up a record by its key.

arnau: I think all collections should use the same pattern and actually I'm interested in getting this collection ordered. Not yet there but this is the first step

content/indexes-removal/index.md
arnau: Well, this the example of the entry I'm proposing to change so it has to be "item".

content/indexes-removal/index.md
arnau: It's a good question, I think it is present because a record is supposed to be an entry, and a record resource is an "enriched" entry with the blob data. I would be happy to revisit this but it's potentially too big for this RFC.

MatMoore: Fair enough, it's a relatively small thing and we could always deprecate it later.

content/indexes-removal/index.md
MatMoore: Ah, I assumed the starting point for this change would be post-RFC16 (https://github.com/openregister/registers-rfcs/pull/29). Doesn't matter as long as it's consistent.

PR #42, opened 2018-10-03
RFC0025: Snapshot resource

arnau:
### Context

One of the most difficult concepts to explain in the Registers ecosystem is
the concept of “record”. The reasons boil down to two things: First, the
concept of “record” has shifted over time as a natural adaptation to how
Registers has evolved and second the current record definition doesn't match
the expectations on the record resource. Probably the second is consequence of
the first.

This RFC proposes creating a new Snapshot resource and redefine the Record
resource to have a single responsibility.

This is a breaking change.

**This is an aggressive take on what the API should be doing** When you review it, focus first on the motivation, if we agree on that, move on to the rest otherwise let's first find common ground on it.

### DIFF
--- a/content/snapshot-resource/index.md
+++ b/content/snapshot-resource/index.md
@@ -1,0 +1,255 @@
+---
+rfc: 0025
+start_date: 2018-10-03
+decision_date:
+pr: openregister/registers-rfcs#42
+status: draft
+---
+
+# Snapshot resource
+
+## Summary
+
+This RFC proposes creating a new Snapshot resource and redefine the Record
+resource to have a single responsibility.
+
+This is a breaking change.
+
+## Motivation
+
+One of the most difficult concepts to explain in the Registers ecosystem is
+the concept of “record”. The reasons boil down to two things: First, the
+concept of “record” has shifted over time as a natural adaptation to how
+Registers has evolved and second the current record definition doesn't match
+the expectations on the record resource. Probably the second is consequence of
+the first.
+
+Let me elaborate on what does “the current record definition doesn't match
+the expectations on the record resource” mean.
+
+The current definition of record is: The most recent entry for a given key.
+The record resource though is the combination of the most recent entry for a
+given key and the blob of data referenced by it.
+
+The reason for having the blob of data merged in is to make the record
+resource useful to the type of user that wants to get the data for a given
+key.
+
+The reason for having the entry data is to provide the context for how it
+relates to the log of entries. Useful to users that need to deal with
+entries, blobs and the log itself.
+
+The first reason is “porcelain” in the sense of “don't make me think, give me
+the data”. The second is “plumbing” in the sense of “let me handle the low
+level bits myself”.
+
+The result is tension between these two needs and an ill-defined concept.
+The proposal is to split both needs into dedicated resources.
+
+## Explanation
+
+Let's define “snapshot” to be the set of most recent entries for each unique key.
+
+```elm
+type Snapshot =
+  Set Entry
+```
+
+The operation to compute the snapshot for a given log can be defined as:
+
+```elm
+collect : Log -> Snapshot
+```
+
+***
+**EXAMPLE:**
+
+For example, given log `log` of size 10, we can compute the latest snapshot
+with:
+
+```elm
+collect log
+```
+
+And compute the snapshot at size 5 by slicing the log to that size:
+
+```elm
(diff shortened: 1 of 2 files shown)


MatMoore:
I think I agree with the motivation.

I'd like a bit more time to think through this RFC, but here are my first thoughts:

- It makes sense to take entry number etc out of the record
- I'm ok with `/records` always returning the latest data. I'm not sure if it will makes sense to parameterise it with log size as I don't know how the user would know what this number means if they are just using `/records` alone
- I think if we allowed custodians to tag entries, that would work well with either kind of resource
- I'm unsure about calling it "snapshot", because the name sounds to me like something that would belong in the porcelain layer, like a snapshot of a record, rather than an entry

content/snapshot-resource/index.md
```diff
@@ -0,0 +1,255 @@
...
+The first reason is “porcelain” in the sense of “don't make me think, give me
+the data”. The second is “plumbing” in the sense of “let me handle the low
+level bits myself”.
+
+The result is tension between these two needs and an ill-defined concept.
+The proposal is to split both needs into dedicated resources.
+
+## Explanation
+
+Let's define “snapshot” to be the set of most recent entries for each unique key.
```
gidsg: I think it would be clearer to say:
*snapshot as a set containing the most recent entry for each key.*

arnau:
@MatMoore, after some consideration

> I'm ok with /records always returning the latest data. I'm not sure if it will makes sense to parameterise it with log size as I don't know how the user would know what this number means if they are just using /records alone

I don't think this is a concern for this RFC in particular but I can see another RFC returning the log size as a HTTP header so that someone consuming the set of records could replicate the exact query at a latter time.

> I think if we allowed custodians to tag entries, that would work well with either kind of resource
I'm unsure about calling it "snapshot", because the name sounds to me like something that would belong in the porcelain layer, like a snapshot of a record, rather than an entry. Something like /entry-snapshots might better signal the relationship between the various endpoints.

I considered this but I don't see "snapshot" having a strong implicit meaning either way (porcelain/plumbing) so I'm still keen to use "snapshot" as defined in the RFC. After all "records" is the porcelain view on the same set offered by "snapshot".

If you still have reservations let's address them in the next RFC corner.

MatMoore:
> I don't think this is a concern for this RFC in particular but I can see another RFC returning the log size as a HTTP header so that someone consuming the set of records could replicate the exact query at a latter time.

That makes sense to me. I think it would work well as a link header, so the user doesn't actually need to know what that `log-size` parameter is, it just functions as a permalink mechanism.


> I considered this but I don't see "snapshot" having a strong implicit meaning either way (porcelain/plumbing) so I'm still keen to use "snapshot" as defined in the RFC. After all "records" is the porcelain view on the same set offered by "snapshot".
> 
Ok, happy to go with snapshot then 👍

content/snapshot-resource/index.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+#### Parameters
+
+|Name|Type|Description|
+|-|-|-|
+|`log-size`|Integer|The log size to compute the snapshot.|
+
+Note: We can consider some kind of alias for the latest log size like the
+`latest` tag used by Docker.
```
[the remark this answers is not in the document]
arnau: Correct.

content/snapshot-resource/index.md
```diff
@@ -0,0 +1,261 @@
...
+
+Returns an array of record resources.
+
+***
+##### Endpoint
+
+```
+GET /records
+```
+
```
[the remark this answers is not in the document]
arnau: The list of records is the full list of records. In terms of the shape of the response, it is an array of records as defined previously.

content/snapshot-resource/index.md
```diff
@@ -0,0 +1,261 @@
...
+Accept: application/json
+Host: country.register.gov.uk
+```
+
+```http
+{
+  "_id": "GB",
+  "name": "United Kingdom",
+  "official-name": "The United Kingdom of Great Britain and Northern Ireland",
+  "citizen-names": ["Briton", "British citizen"]
```
gidsg: this would still have the current register key convention in our usage as well as `_id` right? i.e. `"country": "GB"`

arnau: It would still have the field `country` because it is data part of the blob

PR #43, opened 2018-11-09
RFC0026: Adopting OpenAPI

MatMoore:
### Context
As a developer on the GOV.UK registers team, it's currently a lot of work to keep the registers specification, the conformance tests, the reference implementation, and the API explorer in sync, while at the same time providing high quality reference documentation for the GOV.UK registers service. Manually updating the API reference feels like duplicated effort, especially since we already define schemas for the API in the conformance tests.

When making decisions about the tooling surrounding the reference implementation, I would prefer to automate as much as possible, and reuse existing tools rather than rolling our own.

We should also avoid building too much tooling that is coupled to our own service. In my opinion both the API explorer built into ORJ and the API reference are serving similar needs, but they have been implemented in a way that makes neither suitable for other teams who might want to host the software themselves.

This RFC proposes using OpenAPI to provide a machine readable specification of the JSON REST API. This would allow any implementor to choose existing open source tools for documentation, testing, code generation etc.

### Guidance to review
If this RFC is approved, I suggest I prototype something before merging, as there are a lot of unknowns here.

Another option we could take would be to adopt OpenAPI for internal use first, without committing to providing anything indefinitely.

### DIFF
--- a/content/openapi/index.md
+++ b/content/openapi/index.md
@@ -1,0 +1,73 @@
+---
+rfc: 0026
+start_date: 2018-08-09
+pr: openregisters/registers-rfc#9981
+status: draft
+---
+
+# Use OpenAPI to describe the REST API
+
+## Summary
+
+This RFC proposes that we publish an OpenAPI document for each version of the specification.
+
+The conformance tests for the specification should reuse this rather than defining additional schemas in the code.
+
+Note that at the time of writing OpenAPI only supports JSON and YAML. Therefore the benefits of OpenAPI described in this CSV only apply to implementations that use the JSON representation.
+
+## Motivation
+
+OpenAPI is a standard for describing - in a machine readable format - REST APIs that return JSON or YAML.
+
+In this RFC, when I say OpenAPI I'm refering to [version 3.0.2 of the specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md).
+
+Having a machine readable specification for the registers API allows implementors to take advantage of an existing ecosystem of tooling for building and hosting APIs.
+
+OpenAPI is not the only standard for specifying APIs, but I'm restricting the proposal to OpenAPI because it is a [GDS recommended standard for API documentation](https://www.gov.uk/guidance/gds-api-technical-and-data-standards#document-your-api).
+
+## Explanation
+OpenAPI tools can provide API documentation, generate client libraries, and validate that implementations conform to the specification.
+
+### API documentation
+Currently the reference implementation comes with an API explorer, but no API documentation. The API explorer links out to the GOV.UK Registers API documentation, which is not appropriate for other implementors who might wish to host the software themselves.
+
+Although [the source for this documentation is open](https://github.com/alphagov/registers-tech-docs/), it's not easy to reuse, because
+- the content contains information that is specific to the GOV.UK Registers service
+- the content must be manually updated whenever a change is made to the API (this is also more difficult now that we have multiple API versions)
+
+Providing an OpenAPI document along with the specification would mean that any implementation can reuse it, using their choice of documentation tool.
+
+This RFC does not propose removing the API explorer from the reference implementation, but using a 3rd party documentation tool such as [Swagger UI](https://swagger.io/tools/swagger-ui/) may reduce the need for us to continue to maintain our own solution.
+
+### Conformance testing
+Currently we provide a test suite to verify that an implementation obeys the specification.
+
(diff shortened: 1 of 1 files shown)


MatMoore:
@bahmady do you have any thoughts about this?

gidsg (approved):
Happy in principal to give this a try. From past experience it has been non-trivial to autogenerate swagger depending on the framework so I agree we should spike this to see if it is actually feasible!

bahmady:
I think it'd be a good thing to do, and I'd like to be involved if you'll have me (at least to see what's going on under the hood). 

I think Open API v3 is likely to be more feasible/easy to implement than Swagger/Open API v2. Pay has set up auto-generation of a v2 file. @kbottla or @heathd are generally very knowledgeable about it so might be able to assist with some advice if they have time.

content/openapi/index.md
```diff
@@ -0,0 +1,73 @@
...
+To keep the OpenAPI document implementation-independent, we should focus on the last 3 points, and not specify any server endpoints.
+
+### Discoverability
+The document should be linked from the corresponding specification page.
+
+Each version of the API should have its own OpenAPI document.
+
+### Limitations of OpenAPI
+
+The Registers team has briefly investigated OpenAPI before (see [this comment from Paul Downey](https://github.com/alphagov/open-standards/issues/31#issuecomment-319334209)). A reason against adopting it is that parts of the API (such as HATEOAS, and especially links between registers) would be difficult or impossible to capture. This means that a purely machine generated documentation site would be missing information. This is something we should consider carefully when authoring the document, but I don't consider a strong enough reason to not adopt it.
```
arnau: What is the use case of cross-domain APIs in registers? As far as I can tell the Registers API is based on a single domain (and replicated to as many domains as registers exist).

Interlinking is outside of the actual REST API, done mainly through the CURIE datatype which at this level of definition (HTTP) is treated as an opaque string.

arnau: To clarify on the schema part, OpenAPI has a set of datatypes and validation rules that, at the moment are incompatible with the Registers type system. Friction could be reduced by providing custom datatype formats but it would fail to express properly the flexibility of certain types we have in the registers type system.

For context, a similar issue exists with CSVW, and a similar partial solution can be applied.

My question would be: What is the purpose of providing OpenAPI? machine readable API definition? autogenerated documentation? schema validation reusing tooling from OpenAPI? All of them? Others?

MatMoore: > My question would be: What is the purpose of providing OpenAPI? machine readable API definition? autogenerated documentation? schema validation reusing tooling from OpenAPI? All of them? Others?

If we ignore our own needs for registers.service.gov.uk, I think the main benefit is the option of providing autogenerated documentation out of the box. Anyone who wants their own version of openregister-java needs to host documentation along with it. Things like code generation are a bonus, but it's hard to say whether anyone would use it for that.

The schema validation is more of a convenience for us - if the RFC is adopted I would swap out the JSON schema for the OpenAPI schema in the conformance tests, but the tests would do the same thing either way.

MatMoore: > To clarify on the schema part, OpenAPI has a set of datatypes and validation rules that, at the moment are incompatible with the Registers type system. Friction could be reduced by providing custom datatype formats but it would fail to express properly the flexibility of certain types we have in the registers type system.
> 
> For context, a similar issue exists with CSVW, and a similar partial solution can be applied.
> 

This same issue already exists with the JSON schema we're using in the conformance tests. We treat blobs as arbitrary objects, which means we would need extra code in our python tests to validate them properly (we should probably do this regardless of whether this RFC is accepted).

I'm ok with a partial solution here, because I don't think it prevents autogenerated documentation or client libraries from being useful.

MatMoore: > What is the use case of cross-domain APIs in registers? As far as I can tell the Registers API is based on a single domain (and replicated to as many domains as registers exist).
> 
> Interlinking is outside of the actual REST API, done mainly through the CURIE datatype which at this level of definition (HTTP) is treated as an opaque string.

Yeah, I am happy to treat it opaquely as well. As well as CURIEs in the register we might consider adding links to datatypes in the metadata schema but I this is hypothetical at the moment.

Ideally I'd also like a machine readable spec to cover pagination via link headers. I haven't looked into yet but I expect that's an area where we may run into the limits of what OpenAPI can express.

MatMoore:
Thanks @bahmady. Let me know on slack if there's a good time to chat about this. I'll probably start by converting the JSON schema we have to an OpenAPI 3 spec and see what problems I run into.

content/openapi/index.md
arnau: I would like the RFC to be more precise on what do we expect out of OpenAPI, highlighting the friction points we identified in this thread. With that, I'm happy with the approach.

MatMoore:
I'm going to hand this over to @gidsg as I didn't have time to get this in a good state before my last day at GDS (sorry!)

